PR #1, opened 2023-09-26
Don't bill for projects for certain periods of time

naved001:
This handled mechanism to read a CSV file which describes some projects that shouldn't be billed for a given interval.


naved001:
> First time I have seen pandas used

@joachimweyl  it's my first time using pandas too 😆 Seems pretty powerful so far.

PR #2, opened 2023-09-26
The script now accepts multiple CSV files which it then merges into one

naved001:
dataframe before being processed and written.


naved001:
@joachimweyl I'll move it to MOC i.e. cci-moc

PR #10, opened 2024-03-29
Added ability to group invoices by PIs

### LINKED ISSUE #7: Add post invoice script automation script to break Invoice into groups by PI
## Motivation
Regularly let users know how many resources they have used prior to receiving their official invoice.

## Completion Criteria
Invoice CSV grouped into groups by PI

## Description
- [ ] Create a way to group data in the invoice CSV by PI (individual CSVs or a database whatever is easier, developer's prerogative)
- [ ] This will then be used to email out this information but that is a different issue

## Completion dates
Desired - 2024-02-14
Required - TBD

QuanMPhm:
Closes #7, I have modified `process_report.py` to group the merged invoices into ones grouped by PIs. These smaller CSVs will be placed in a folder of your choice, or defaults to a one placed in the working directory

### DIFF
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -1,4 +1,5 @@
 import argparse
+import os
 
 import pandas
 
@@ -34,6 +35,12 @@ def main():
         default="filtered_output.csv",
         help="Name of output file",
     )
+    parser.add_argument(
+        "--output-folder",
+        required=False,
+        default="pi_invoices",
+        help="Name of output folder containing pi-specific invoice csvs"
+    )
     args = parser.parse_args()
     merged_dataframe = merge_csv(args.csv_files)
 
@@ -53,8 +60,9 @@ def main():
 
     projects = list(set(projects + timed_projects_list))
 
-    remove_non_billables(merged_dataframe, pi, projects, args.output_file)
+    billable_projects = remove_non_billables(merged_dataframe, pi, projects, args.output_file)
     remove_billables(merged_dataframe, pi, projects, "non_billable.csv")
+    export_pi_billables(billable_projects, args.output_folder)
 
 
 def merge_csv(files):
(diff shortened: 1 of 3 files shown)


process_report/process_report.py
```diff
@@ -53,8 +60,9 @@ def main():
 
     projects = list(set(projects + timed_projects_list))
 
-    remove_non_billables(merged_dataframe, pi, projects, args.output_file)
+    billable_projects = remove_non_billables(merged_dataframe, pi, projects, args.output_file)
```
naved001, line 63: can we still print the whole billable csv?

process_report/process_report.py
```diff
@@ -53,8 +60,9 @@ def main():
...
     remove_billables(merged_dataframe, pi, projects, "non_billable.csv")
+    export_pi_billables(billable_projects, args.output_folder)
```
naved001, line 65: can we also make this optional and triggered by a flag? I think the default behavior should still be to produce the 2 reports i.e. billable and non-billable, but the user can say `--export-by-pi` and then it should split it out into multiple files. @joachimweyl thoughts?

process_report/process_report.py
QuanMPhm: @naved001 I only modified `remove_non_billable` to return the dataframe with billable projects, so the function should still write out the billable csv

naved001: oh, right. I missed it, thanks!

process_report/process_report.py
joachimweyl: that makes sense, as long as it is a flag we know to use and use every month until we decide to be done with it.

naved001: @QuanMPhm pointed out that the billable csv is still written, so I am okay with the current behavior then and don't care much about the flag.

QuanMPhm: @naved001 @joachimweyl It should only take minutes for me to make it optional for the script to export the invoice csv by pi or as combined or as both. I'll do what you prefer.

naved001 (approved):

PR #11, opened 2024-04-01
Amending commit to add ability to group invoices by PIs

QuanMPhm:
I believe the bug comes from two lines which tries to index into the dataframe. I used an inappropriate method for indexing and this should fix it.

### DIFF
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -119,12 +119,12 @@ def export_pi_billables(dataframe: pandas.DataFrame, output_folder):
     if not os.path.exists(output_folder):
         os.mkdir(output_folder)
 
-    invoice_month = dataframe['Invoice Month'][0]
+    invoice_month = dataframe['Invoice Month'].iat[0]
     pi_list = dataframe['Manager (PI)'].unique()
 
     for pi in pi_list:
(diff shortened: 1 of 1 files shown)


process_report/process_report.py
```diff
@@ -119,12 +119,12 @@ def export_pi_billables(dataframe: pandas.DataFrame, output_folder):
     if not os.path.exists(output_folder):
         os.mkdir(output_folder)
 
-    invoice_month = dataframe['Invoice Month'][0]
+    invoice_month = dataframe['Invoice Month'].iat[0]
```
QuanMPhm, line 122: @naved001 The intent of this line and line 127 was to quickly obtain the invoice month and the PI's instituition name from the dataframe. I tried to get this information by indexing into the first element of the `Invoice Month` and `Institution` column.
Using `[]` for indexing was a form of label indexing similar to [`.loc[]`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.loc.html#), which would raise an error if row 0 of the invoice dataframe was not present (which would happen if the project in the first row was unbillable/excluded). `iat[]` uses [position indexing,](https://pandas.pydata.org/docs/reference/api/pandas.Series.iat.html) which would be the appropriate method for this case.

naved001 (approved):

PR #12, opened 2024-04-03
Added more processing logic to find institution name, applying credits, and to export HU and BU invoices

### LINKED ISSUE #6: Automate More of the Invoice post processing steps
## Motivation
After the OpenShift, OpenStack, and storage invoice data is pulled it is then combined together. Then the next step is that it is processed in a few ways in Google Sheets. We would like to automate this after processing as much of the last step as we can easily.
We are not trying to automate all of the steps, just the steps leading up to what we send to MGHPCC. The other steps such as updating PDF's, Rollups, Orran course charges, and Utilized billed and non billed can wait as @joachimweyl has automated them mostly in Google Sheets and they are not on the same time crunch.

## Completion Criteria
Provide details of the steps of the Invoice after-process

## Description
1. Add 3 new columns, Credit, Credit Code, and Balance to billable CSV
   1. Add Credits for new PI
      1. Compare a list of all PIs that have used OpenStack and OpenShift in the past, find out if there are any new PIs this month, and if there are add the current Cost value (up to $1k across the whole project) to the Credit column of the billable CSV 
      1. this means that if there are 4 projects and they sum up to 1.5k we only add up to 1k in the Credit column and the rest will be invoiced directly
   1. Add 0002 to the Credit code field for all rows that are given New PI Credits
   1. Set the Balance column for all rows to be Cost minus Credit
      1. this means for non-credit rows balance will equal cost and for credit rows it will be the difference.
1. Update the Institution column
   1. You can use the following excel code to convert, knowing that this can change we might want this to be a table in the [invoicing repo](https://github.com/nerc-project/operations/issues/420). Keep in mind the code below is order-specific because of Harvard affiliates.
   ```
   =IF(REGEXMATCH(D2,"northeastern.edu"),"Northeastern University",IF(REGEXMATCH(D2,"bu.edu"),"Boston University",IF(REGEXMATCH(D2,"bentley.edu"),"Bentley",IF(REGEXMATCH(D2,"uri.edu"),"University of Rhode Island",IF(REGEXMATCH(D2,"redhat.com"),"Red Hat",IF(REGEXMATCH(D2,"childrens.harvard.edu"),"Boston Childrens Hospital",IF(REGEXMATCH(D2,"mclean.harvard.edu"),"McLean Hospital",IF(REGEXMATCH(D2,"meei.harvard.edu"),"Massachusetts Eye & Ear",IF(REGEXMATCH(D2,"dfci.harvard.edu"),"Dana-Farber Cancer Institute",IF(REGEXMATCH(D2,"bwh.harvard.edu"),"Brigham and Women's Hospital",IF(REGEXMATCH(D2,"bidmc.harvard.edu"),"Beth Israel Deaconess Medical Center",IF(REGEXMATCH(D2,"harvard.edu"),"Harvard University",IF(REGEXMATCH(D2,"wpi.edu"),"Worcester Polytechnic Institute",IF(REGEXMATCH(D2,"mit.edu"),"Massachusetts Institute of Technology",IF(REGEXMATCH(D2,"umass.edu"),"University of Massachusetts Amherst",IF(REGEXMATCH(D2,"uml.edu"),"University of Massachusetts Lowell",IF(REGEXMATCH(D2,"codeforboston.org"),"Code For Boston",IF(REGEXMATCH(D2,"mmsh"),"Harvard University",IF(REGEXMATCH(D2,"gstuart"),"University of Massachusetts Amherst",IF(REGEXMATCH(D2,"rudolph"),"Boston Childrens Hospital",IF(REGEXMATCH(D2,"robbaron"),"Boston University",IF(REGEXMATCH(D2,"kmdalton"),"Harvard University",IF(REGEXMATCH(D2,"mzink"),"University of Massachusetts Amherst",IF(REGEXMATCH(D2,"@yale.edu"),"Yale University",IF(REGEXMATCH(D2,"francesco.pontiggia"),"Harvard University","Please add new Institution to function")))))))))))))))))))))))))
   ```

1. Create a CSV of only the HU rows named NERC-YYYY-MM-Harvard-Invoice.csv where YYYY is the year and MM is the month
1. create a CSV of HU and BU only named NERC <Month> Invoice - HU&BU.csv where <Month> is the name of the month of the data processed.
1. Save 2 CSVs above to the [CSV folder](https://drive.google.com/drive/folders/1ybuc7hTSP5CkxQNQ5HkBsC_MZcjXVGYC?usp=drive_link) or to a shared drive that @joachimweyl has access to

## Reference
1. [February data](https://docs.google.com/spreadsheets/d/1ZZELpyIpmcmCCwZXt3wiccKke8ljyW_WFOyd1NnMIVY/edit?usp=sharing)

## Completion dates
Desired - 2024-01-25
Required - TBD

QuanMPhm:
Closes #6 This is a draft PR that implemented most of the functionalities @joachimweyl asked for. These include:
1. `add_institution()` to get the institute name for all PIs in the invoice
2. `apply_credits_new_pi()` to apply the New PI credit to all projects in the invoice
3. `export_HU_only()` and `export_HU_BU()` to export HU and BU specific invoices

A feature worth considering is to export the CSVs to a storage service like Google Drive, will per @knikolla's suggestion, should be put in a separate issue.

Aside from the implementation of the processing functions to apply credits and institution, as well as the export functions, some structural changes are added: 

1. The invoice fields names have been placed in top-level constants to provide a single convenient place to edit these field names, should they change in the future
5. The mapping between institution email domains and institution name is moved to the configuration file `institute_map.json`


process_report/process_report.py
```diff
@@ -1,9 +1,106 @@
...
+        "bentley.edu"           : "Bentley",
+        "uri.edu"               : "University of Rhode Island",
+        "redhat.com"            : "Red Hat",
+        "childrens.harvard.edu" : "Boston Childrens Hospital",
+        "mclean.harvard.edu"    : "McLean Hospital",
+        "meei.harvard.edu"      : "Massachusetts Eye & Ear",
+        "dfci.harvard.edu"      : "Dana-Farber Cancer Institute",
+        "bwh.harvard.edu"       : "Brigham and Women's Hospital",
+        "bidmc.harvard.edu"     : "Beth Israel Deaconess Medical Center",
+        "harvard.edu"           : "Harvard University",
```
joachimweyl, line 81: I do not know how Python prioritizes matches, since any of the 6 items above this line could match to this row how do we make sure that any of the 6 above match their value, and then only harverad.edu matches not in the list above becomes Harvard University?

QuanMPhm: @joachimweyl From what I found, Python dictionary is [ordered](https://docs.python.org/3/library/stdtypes.html#dict.values), so the row will be matched against the Harvard affiliates before being matched against `harvard.edu`. I added this scenario in my [test case ](https://github.com/CCI-MOC/process_csv_report/pull/12/files#diff-2ab8a00ce6228ed90b8929a4e45622b1586da07acbc85aa2d71533747f594ca9R188) to make sure.

process_report/process_report.py
```diff
@@ -1,9 +1,106 @@
...
+    
+    return dataframe
+
+
+applied_credits = [apply_credits_0002]
+
+
+def get_institution_from_pi(pi_uname):
+    
+    institute_map = {
```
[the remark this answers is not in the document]
QuanMPhm: @joachimweyl I have no preference either way. @naved001 Thoughts?

process_report/process_report.py
```diff
@@ -1,9 +1,106 @@
...
 
+### Invoice field names
+PROJECT_FIELD = 'Project - Allocation'
+MANAGER_PI_FIELD = 'Manager (PI)'
+INSTITUTION_FIELD = 'Institution'
+COST_FIELD = 'Cost'
+CREDIT_FIELD = 'Credit'
+CREDIT_CODE_FIELD = 'Credit Code'
+BALANCE_FIELD = 'Balance'
+###
```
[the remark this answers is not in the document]
joachimweyl: @naved001 you created this original code, do you have a preference?

process_report/tests/unit_tests.py
```diff
@@ -175,3 +176,65 @@ def test_export_pi(self):
...
+class TestCredit0002(TestCase):
+    def setUp(self):
+
+        data = {
+            'Manager (PI)': ['PI1', 'PI1', 'PI2', 'PI3', 'PI4', 'PI4'],
+            'Project - Allocation': ['ProjectA', 'ProjectB', 'ProjectC', 'ProjectD', 'ProjectE', 'ProjectF'],
+            'Cost': [10, 100, 10000, 5000, 800, 1000]
+        }
+        self.dataframe = pandas.DataFrame(data)
+        old_pi = ['PI2', 'PI3']
```
joachimweyl, line 207: I am probably missing something. do we fill the old_pi file with previous values so we can test that it subtracts from this value? I also want to make sure that it only is providing this discount for 2 months no more. In other words, as soon as a PI has been on the list for 2 months even if they have not gone through the $1000 they don't get any more.

process_report/tests/unit_tests.py
```diff
@@ -175,3 +176,65 @@ def test_export_pi(self):
...
+            self.assertEqual(institution_name, row['Answer'])
+
+@skipIf(process_report.apply_credits_0002 not in process_report.applied_credits, "Skipping test for credit 0002 because credit not enabled")
+class TestCredit0002(TestCase):
+    def setUp(self):
+
+        data = {
+            'Manager (PI)': ['PI1', 'PI1', 'PI2', 'PI3', 'PI4', 'PI4'],
+            'Project - Allocation': ['ProjectA', 'ProjectB', 'ProjectC', 'ProjectD', 'ProjectE', 'ProjectF'],
+            'Cost': [10, 100, 10000, 5000, 800, 1000]
```
joachimweyl, line 204: is there a reason you did not include and answer key here?

joachimweyl:
I don't know Python well enough to check your code but the logic is sound.

process_report/process_report.py
```diff
@@ -1,9 +1,106 @@
...
+    except Exception:
+        print("Old PI file does not exist, or the C0002_OLD_PI env var is not set")
+        sys.exit(1)
+
+    print("Old pi list: ", old_pi_list)
+    pi_list = dataframe[MANAGER_PI_FIELD].unique()
+
+    for pi in pi_list:
+        if pi != pi: continue   # NaN check
+        if pi in old_pi_list: continue  # Is the PI an old PI?
```
joachimweyl, line 44: If this code is run for testing will the count of months the PI has been on the old PI list go up or are we actually tracking months?

process_report/tests/unit_tests.py
QuanMPhm: @joachimweyl The way I have implemented for the "New PI" credit is applied is that:
You provide a file containing a list of old PIs (PIs who will not receive the "New PI" credit in this invoice). You must create and maintain this "Old PI file" yourself. My code so far doesn't contain any logic about how long the PI has joined NERC or etc. It currently does not get information from any external data sources, other than the files the person running the script has provided.

process_report/tests/unit_tests.py
QuanMPhm: @QuanMPhm The assertions made here check to see if the correct credit and balances were given to the PIs. I could have put the answers for the credits and balances in that dict, but that would just require more code to reference those answers in the test function `test_apply_credit_0002()`.

process_report/process_report.py
QuanMPhm: @joachimweyl My code has no logic to track how long a PI has appeared on an invoice.

process_report/tests/unit_tests.py
joachimweyl: to automate the credit code 0002 we need to:
1) have an existing PI list (new_pi.csv) that has 2 columns 1 is the pi address and 2 is age (filled with 1, 2, old) or something similar
   1) this could be combined with a new column that tracks how much of the 1000 they have used up
2) update the  list based on the current data file
3) provide credit based on the second column

process_report/process_report.py
joachimweyl: see discussion in unit_tests.py line 207

process_report/process_report.py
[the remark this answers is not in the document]
naved001: @QuanMPhm let's add them for all the columns

process_report/tests/unit_tests.py
QuanMPhm: @joachimweyl The creation of a `new_pi.csv` would be a seperate problem. With your permission, I could download all the invoice data we have since as far back as you'd like, and write a script to generate the `new_pi.csv`. After that I can include logic to update `new_pi.csv`.

I assume this probably mean `new_pi.csv` has to be version controlled and put in this repo? And we'd have to submit a monthly PR to update the one file. I not exactly certain if this is the best way to do thing, and would rather have opinions from @knikolla and @naved001.

process_report/process_report.py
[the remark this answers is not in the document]
naved001: external file sounds better actually.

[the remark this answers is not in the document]
QuanMPhm: I've moved the mapping into the external file

process_report/process_report.py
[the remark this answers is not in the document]
QuanMPhm: I have made a constant for all the columns

process_report/tests/unit_tests.py
joachimweyl: agreed, I would rather this automatically update when the script is run, not requiring a PR to update if possible.

process_report/process_report.py
```diff
@@ -1,9 +1,97 @@
...
+BALANCE_FIELD = 'Balance'
+###
+
+
+def apply_credits_0001(dataframe):
+    credit_code = "0001"
+    pass
+
+
+def apply_credits_0002(dataframe):
```
knikolla, line 32: I would rather this be called `apply_credits_new_pis`.

process_report/process_report.py
```diff
@@ -1,9 +1,97 @@
...
+            continue   # NaN check
+        if pi in old_pi_list: 
+            continue  # Is the PI an old PI?
+
+        pi_projects = dataframe[dataframe[PI_FIELD] == pi]
+        rem_credit = credit_amount
+        for i, row in pi_projects.iterrows():
+            project_cost = row[COST_FIELD]
+            if project_cost >= rem_credit:
+                dataframe.at[i, CREDIT_FIELD] =  rem_credit
```
knikolla, line 65: @joachimweyl @QuanMPhm I am slightly confused by this. Is our desired outcome that for a PI that has multiple projects and a credit, the credit field in the CSV decrease row by row as the credits are exhausted or do we want the credit field in all rows to be the final computation?

knikolla (changes requested):
Thanks for proposing this PR and the good work.

Some suggestions, besides the comments inline. This PR implements at least 3 new features to this codebase.

- Deriving institution name from a PIs username.
- Exporting two additional versions of the output for only BU, and HU/BU.
- The credits system.

There's not a lot of overlap between them and in situations like this in the future I would encourage you to break them up into several PRs. In this case I would have gone with 3. 

This makes the code easier to review. You don't need to break this one up, just a suggestion for the future.

process_report/process_report.py
joachimweyl: this is based on a Google sheet that the easiest way to do it was to decrease row by row, that being said it does not matter how it is done as long as we reduce by $100 or total whichever is smaller.

knikolla: @joachimweyl $100?

joachimweyl: sorry $1000, missed a zero.

knikolla: Then I would strongly recommend not decreasing the credit amount line by line but making sure all cells for a PI are correct.

process_report/process_report.py
```diff
@@ -1,9 +1,97 @@
...
+
+
+def apply_credits_0001(dataframe):
+    credit_code = "0001"
+    pass
+
+
+def apply_credits_0002(dataframe):
+    """Applies the New PI Credit. This credit function expects the 
+    env var `C0002_OLD_PI` to be set, pointing to a txt file containing old PIs"""
```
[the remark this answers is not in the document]
QuanMPhm: @knikolla When deciding how to pass arguments to the credit functions, I thought that there wasn't a simple way without using env vars. My thought process went like this:

- There is the assumption that we will have use and discontinue more credits in the future. These credits may require different parameters of varying amounts.  
- This assumption influenced how I designed the credit functions and `add_credits()`. The credit functions only take the invoice dataframe as the only function argument, which would allow `add_credits()` to simply iterate through the credit functions, instead of having to make an `if` block for each credit function, then pass in its specific parameters, like this:
```
def add_credits(dataframe, args):

    # Insert rows
    dataframe.insert(dataframe.columns.get_loc(COST_FIELD) + 1, CREDIT_FIELD, 0.0)
    dataframe.insert(dataframe.columns.get_loc(CREDIT_FIELD) + 1, CREDIT_CODE_FIELD, None)
    dataframe.insert(dataframe.columns.get_loc(CREDIT_CODE_FIELD) + 1, BALANCE_FIELD, 0.0)

    # Apply credits
    if credit_is_active(0001):
        dataframe = apply_credits_0001(dataframe, args.C0001)
    if credit_is_active(0002):
        dataframe = apply_credits_0002(dataframe, args.C0002_old_pi)
    if credit_is_active(0003):
        ...
```
- I thought that as more credits come and go, writing `add_credits` like this would just add another section of `process_report.py` that needs to be continually updated. Although not much work would be needed to update this function, it is still work nonetheless. 
- Given all this, I decided to pick the first approach. With all credit functions taking one arguement, the credit-specific parameters would have to come from env vars.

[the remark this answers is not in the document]
knikolla: I understand your reasoning, however:

- It doesn't have to be an argument to the function, there are ways to avoid that. Example: global variables, configuration options accessible globally. My ask here was for this to be an argument to the command line as this makes the application much simpler to use as opposed to an environment variable.
- In general, my preference is to try to avoid creating an abstraction or generalization when you have only one case to work with. We don't know what may extra information will be required to apply other kinds of credits and we will not have any other kinds of credits in the near term. It is extremely likely that whatever generalization you come up with here will not work for the second case of credits anyway. Start simple.

process_report/process_report.py
```diff
@@ -1,9 +1,97 @@
...
+SU_HOURS_FIELD = 'SU Hours (GBhr or SUhr)'
+SU_TYPE_FIELD = 'SU Type'
+COST_FIELD = 'Cost'
+CREDIT_FIELD = 'Credit'
+CREDIT_CODE_FIELD = 'Credit Code'
+BALANCE_FIELD = 'Balance'
+###
+
+
+def apply_credits_0001(dataframe):
```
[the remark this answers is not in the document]
QuanMPhm: @knikolla I assumed we would need to implement the credit function for the 0001 credit at some point, so decided to place a stub there. I can remove it if you think it is unnecessary.

process_report/process_report.py
```diff
@@ -1,9 +1,97 @@
...
+    credit_code = "0001"
+    pass
+
+
+def apply_credits_0002(dataframe):
+    """Applies the New PI Credit. This credit function expects the 
+    env var `C0002_OLD_PI` to be set, pointing to a txt file containing old PIs"""
+    credit_code = "0002"
+    credit_amount = 1000
```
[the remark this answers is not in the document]
QuanMPhm: @knikolla Doesn't the function and surrounding lines give enough context for the `credit_code` and `credit_amount` variable?

process_report/process_report.py
knikolla: The code block confused me as it seem to save applied credit to the cell, rather than remaining credit. 

I suggested Quan to rewrite it to remove the if-else logic and use something like 
```python
>>> applied_credit = min(project_cost, remaining_credit)
>>> remaining_credit -= applied_credit
```

process_report/process_report.py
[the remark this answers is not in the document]
knikolla: The comment does. The name of the function and the name of the variables don't.

Do not depend on the surrounding lines of code or comments as you are not guaranteed that the surrounding lines are going to remain the same as the codebase evolves. Whenever I can avoid a comment entirely by naming the function and variables appropriately, I do.

process_report/process_report.py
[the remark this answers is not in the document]
QuanMPhm: I've restructured the code and added `old-pi-file` as a command arguement

process_report/process_report.py
```diff
@@ -1,9 +1,97 @@
...
+
+
+def get_institution_from_pi(pi_uname):
+    
+    dir_path = os.path.dirname(__file__)
+    with open(f'{dir_path}/institute_map.json', 'r') as f:
+        institute_map = json.load(f)
+
+    for name, institute in institute_map.items():
+        if name in pi_uname:
```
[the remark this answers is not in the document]
QuanMPhm: I've added a docstring for the `add_institution()` function explaining the mapping file

process_report/process_report.py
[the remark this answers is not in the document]
QuanMPhm: Based on your other feedback, I have removed the stub function.

process_report/process_report.py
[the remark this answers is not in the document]
QuanMPhm: I've renamed the variables in the function.

process_report/process_report.py
QuanMPhm: I renamed it to `apply_credits_new_pi`. I felt "New PI credit" sounded more right than "New PIs credit".

process_report/process_report.py
QuanMPhm: The code block has been rewritten.

process_report/process_report.py
```diff
@@ -112,21 +216,60 @@ def remove_billables(dataframe, pi, projects, output_file):
...
+        if pandas.isna(pi): 
+            continue
+        pi_projects = dataframe[dataframe[PI_FIELD] == pi]
+        pi_instituition = pi_projects[INSTITUTION_FIELD].iat[0]
         pi_projects.to_csv(output_folder + f"/{pi_instituition}_{pi}_{invoice_month}.csv")
-        
+
+
+def add_credits(dataframe : pandas.DataFrame):
+    """Adds credits to PIs depending on different criterions"""
```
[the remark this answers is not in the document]
QuanMPhm: The function has been removed, so my typo no longer exists

QuanMPhm:
The ordering of `institute_map.json` no longer matter, as we are using exact matches instead of substrings. Per @naved001's [suggestion](https://github.com/CCI-MOC/process_csv_report/pull/14#discussion_r1558166557).

process_report/process_report.py
```diff
@@ -112,21 +172,99 @@ def remove_billables(dataframe, pi, projects, output_file):
...
+    except FileNotFoundError:
+        print("Applying credit 0002 failed. Old PI file does not exist")
+        sys.exit(1)
+
+    pi_list = dataframe[PI_FIELD].unique()
+
+    for pi in pi_list:
+        if pandas.isna(pi): 
+            continue   # NaN check
+        if pi in old_pi_list:
```
knikolla, line 216: This function does too much. I would split this out into its own function `is_new_pi` that returns a bool whether the PI is old or new. Move over the logic for the file reading, etc into that too.

process_report/process_report.py
```diff
@@ -112,21 +172,99 @@ def remove_billables(dataframe, pi, projects, output_file):
...
+
+    except FileNotFoundError:
+        print("Applying credit 0002 failed. Old PI file does not exist")
+        sys.exit(1)
+
+    pi_list = dataframe[PI_FIELD].unique()
+
+    for pi in pi_list:
+        if pandas.isna(pi): 
+            continue   # NaN check
```
knikolla, line 215: You should at the very least log a scary warning here, by now we should have removed the non-billable projects and therefore if there's something without a PI it should be logged because it's not expected.

knikolla (changes requested):

process_report/process_report.py
QuanMPhm: You want the file reading and the pi checking to be bundled in one function? I'm having a hard time to see how to do that.

knikolla: Two functions then. You can have a function `load_old_pis(old_pi_file) -> dict()` and `is_new_pi(str: username, str: invoice_month) -> bool`

knikolla: Additionally, if we want to be able to run this program reliably, the PI file will need to contain the first invoice month of a PI, otherwise the execution of the program will not be idempotent in the case where this program will update the PI file as it executes.

QuanMPhm: Ok, I'll make it so that PIs whose first invoice month is the same as the month of the invoice being processed will be counted as new PIs for the purpose of applying the credit.

process_report/process_report.py
QuanMPhm: @knikolla I can make a function that validates the dataframe after we remove non-billable projects. For now, I'll only check if the PI name field is empty and log it. Is that fine for now?

process_report/process_report.py
```diff
@@ -112,21 +194,107 @@ def remove_billables(dataframe, pi, projects, output_file):
...
+    Exact matches are then mapped to the corresponding institution name. 
+
+    I.e "foo@bu.edu" would match with "bu.edu", which maps to the instition name "Boston University"
+
+    The list of mappings are defined in `institute_map.json`.
+    """
+    for i, row in dataframe.iterrows():
+        pi_name = row[PI_FIELD]
+        if pandas.isna(pi_name): 
+            print(f"Project {row[PROJECT_FIELD]} has no PI") # Nan check
```
knikolla, line 281: This comment is unnecessary because the function name describes the same thing due to its name `isna`.

process_report/process_report.py
```diff
@@ -1,9 +1,68 @@
...
+###
+
+
+def get_institution_from_pi(pi_uname):
+    
+    dir_path = os.path.dirname(__file__)
+    with open(f'{dir_path}/institute_map.json', 'r') as f:
+        institute_map = json.load(f)
+
+    if '@' in pi_uname:
```
knikolla, line 33: This can be rewritten as the following, due to how split will still work regardless of whether there is `@` in the string, and how we always want the last entry in the list, accessible by `-1` index.

```python
institution_key = pi_uname.split('@')[-1]
institution = institute_map.get(institution_key, '')
```

knikolla (changes requested):

process_report/process_report.py
```diff
@@ -112,21 +194,107 @@ def remove_billables(dataframe, pi, projects, output_file):
...
+            for i, row in pi_projects.iterrows():
+                project_cost = row[COST_FIELD]
+                applied_credit = min(project_cost, remaining_credit)
+
+                dataframe.at[i, CREDIT_FIELD] =  applied_credit
+                dataframe.at[i, CREDIT_CODE_FIELD] = new_pi_credit_code
+                dataframe.at[i, BALANCE_FIELD] = row[COST_FIELD] - applied_credit
+                remaining_credit -= applied_credit
+
+                if remaining_credit == 0:
```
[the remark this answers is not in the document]
QuanMPhm: I figured the break statement would just help save a bit more time. We could break early instead of iterating through all the projects when all the credits have been spent.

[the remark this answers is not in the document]
knikolla: But we still want to populate those Credits fields with 0, no?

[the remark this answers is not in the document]
QuanMPhm: From the [invoices](https://docs.google.com/spreadsheets/d/1O7Kqu0cF2nVqNTuNWdzl9NXX6Gg5eB4IZW4LQDItU3w/edit#gid=1189051626) @joachimweyl shared with me, its seems it's fine to leave the credit field empty if no credit is applied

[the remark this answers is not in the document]
knikolla: Alright.

process_report/process_report.py
```diff
@@ -1,9 +1,67 @@
...
+                old_pi_dict[pi] = first_month
+    except FileNotFoundError:
+        print('Applying credit 0002 failed. Old PI file does not exist')
+        sys.exit(1)
+    
+    return old_pi_dict
+
+
+def is_old_pi(old_pi_dict, pi, invoice_month):
+    if pi in old_pi_dict and old_pi_dict[pi] != invoice_month:
```
knikolla, line 60: This will not work if you are processing old invoices. Say you want to process invoices for `2024-02` that is 2 months ago. A PI with first invoice of `2024-03` is going to appear as an old PI.

I'm okay with fixing this in a follow-up PR while introducing a mechanism to update the PI file.

knikolla (approved):
This looks okay, though I'd like @naved001 to review as well before merging in case I missed something.

process_report/tests/unit_tests.py
```diff
@@ -175,3 +179,97 @@ def test_export_pi(self):
...
+        data = {
+            'Invoice Month': ['2024-03','2024-03','2024-03','2024-03','2024-03','2024-03'],
+            'Manager (PI)': ['PI1', 'PI1', 'PI2', 'PI3', 'PI4', 'PI4'],
+            'Project - Allocation': ['ProjectA', 'ProjectB', 'ProjectC', 'ProjectD', 'ProjectE', 'ProjectF'],
+            'Cost': [10, 100, 10000, 5000, 800, 1000]
+        }
+        self.dataframe = pandas.DataFrame(data)
+        old_pi = ['PI2,2023-09', 'PI3,2024-02', 'PI4,2024-03'] # Case with old and new pi in pi file
+        old_pi_file = tempfile.NamedTemporaryFile(delete=False, mode='w', suffix='.csv')
+        for pi in old_pi: old_pi_file.write(pi + "\n")
```
naved001, line 226: please write this in 2 lines

naved001: @QuanMPhm after this is addressed you can merge this PR!

PR #14, opened 2024-04-04
Added processing and exporting for Lenovo SU Types

### LINKED ISSUE #13: Update Invoice combination script to also generate a Lenovo Loan only CSV
## Motivation
We need to provide Lenovo with a usage list of how much their loaned nodes are being used

## Completion Criteria
A separate CSV is exported to provide Lenovo GPU usage only.

## Description
- [x] Add new [SU types OpenShift GPUA100SXM4](https://github.com/CCI-MOC/openshift-usage-scripts/issues/48)
- [x] Add new [SU types OpenStack GPUA100SXM4](https://github.com/CCI-MOC/openstack-billing-from-db/issues/47)
- [ ] Split out a new Lenovo_YYYY_MM.csv that includes only the OpenShift GPUA100SXM4 & OpenStack  GPUA100SXM4 SUs
   - [Example](https://docs.google.com/spreadsheets/d/1O7Kqu0cF2nVqNTuNWdzl9NXX6Gg5eB4IZW4LQDItU3w/edit?pli=1#gid=179804284)
   - Include only these columns
      - `Invoice Month`
      - `Project - Allocation`
      - `Institution`
      - `SU Hours (GBhr or SUhr)` - rename to `SU Hours`
      - `SU Type`
   - add 2 new columns
      - `SU Charge` - fill completely with $1.00
         - Worth making this a variable since it will change over time
      - `Charge` - Calculate by multiplying SU Hours by SU Charge
- [ ] Test that Lenovo CSV includes only Lenovo loaned GPU Usage

## Completion dates
Desired - 2024-03-27
Required - 2024-04-18

QuanMPhm:
Closes #13, I have added an export function `export_lenovo()` which exports an invoice CSV following the specifications in the related issue. The list of SU types considered to be Lenovo and the SU Charge of $1 have been put into constant values defined in the function.

This PR is blocked by #12.

### DIFF
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -145,6 +145,7 @@ def main():
     export_pi_billables(billable_projects, args.output_folder)
     export_HU_only(billable_projects, args.HU_invoice_file)
     export_HU_BU(billable_projects, args.HU_BU_invoice_file)
+    export_lenovo(billable_projects)
 
 
 def merge_csv(files):
@@ -293,6 +294,26 @@ def export_HU_BU(dataframe, output_file):
     HU_BU_projects = dataframe[(dataframe[INSTITUTION_FIELD] == 'Harvard University') | 
                                (dataframe[INSTITUTION_FIELD] == 'Boston University')]
     HU_BU_projects.to_csv(output_file)        
+
+
+def export_lenovo(dataframe: pandas.DataFrame, output_file=None):
+
+    lenovo_file_name = output_file or f'Lenovo_{dataframe[INVOICE_DATE_FIELD].iat[0]}.csv'
+
+    LENOVO_SU_TYPES = ['OpenShift GPUA100SXM4', 'OpenStack GPUA100SXM4']
+    SU_CHARGE_MULTIPLIER = 1
+
+    lenovo_df = dataframe[dataframe[SU_TYPE_FIELD].isin(LENOVO_SU_TYPES)][[
+            INVOICE_DATE_FIELD, 
+            PROJECT_FIELD, 
+            INSTITUTION_FIELD, 
+            SU_HOURS_FIELD, 
+            SU_TYPE_FIELD]]
+    
+    lenovo_df.rename(columns={SU_HOURS_FIELD: 'SU Hours'}, inplace=True)
+    lenovo_df.insert(len(lenovo_df.columns), 'SU Charge', SU_CHARGE_MULTIPLIER)
+    lenovo_df['Charge'] = lenovo_df['SU Hours'] * lenovo_df['SU Charge']
+    lenovo_df.to_csv(lenovo_file_name)
 
 
 if __name__ == "__main__":
--- a/process_report/tests/unit_tests.py
+++ b/process_report/tests/unit_tests.py
@@ -274,3 +274,41 @@ class TestValidateBillables(TestCase):
         self.assertEqual(1, len(self.dataframe[pandas.isna(self.dataframe['Manager (PI)'])]))
         validated_df = process_report.validate_pi_names(self.dataframe)
         self.assertEqual(0, len(validated_df[pandas.isna(validated_df['Manager (PI)'])]))
+
+
+class TestExportLenovo(TestCase):
+    def setUp(self):
+
+        data = {
+            'Invoice Month': ['2023-01','2023-01','2023-01','2023-01','2023-01', '2023-01'],
+            'Project - Allocation': ['ProjectA', 'ProjectB', 'ProjectC', 'ProjectD', 'ProjectE', 'ProjectF'],
+            'Institution': ['A', 'B', 'C', 'D', 'E', 'F'],
+            'SU Hours (GBhr or SUhr)': [1, 10, 100, 4, 432, 10],
+            'SU Type': ['OpenShift GPUA100SXM4', 'OpenShift GPUA100', 'OpenShift GPUA100SXM4', 'OpenStack GPUA100SXM4', 'OpenStack CPU', 'OpenStack GPUK80']
+        }
+        self.dataframe = pandas.DataFrame(data)
+
+        output_file = tempfile.NamedTemporaryFile(delete=False, mode='w', suffix='.csv')
+        self.output_file = output_file.name
+
+    def tearDown(self):
+        os.remove(self.output_file)
+
+
+    def test_apply_credit_0002(self):
+        process_report.export_lenovo(self.dataframe, self.output_file)
+        output_df = pandas.read_csv(self.output_file)
+
+        self.assertTrue(set([
+            process_report.INVOICE_DATE_FIELD, 
+            process_report.PROJECT_FIELD, 
+            process_report.INSTITUTION_FIELD, 
+            process_report.SU_TYPE_FIELD,
+            'SU Hours', 
+            'SU Charge',
+            'Charge',
+        ]).issubset(output_df))
+        
+        for i, row in output_df.iterrows():
+            self.assertIn(row[process_report.SU_TYPE_FIELD], ['OpenShift GPUA100SXM4', 'OpenStack GPUA100SXM4'])
+            self.assertEqual(row['Charge'], row['SU Charge'] * row['SU Hours'])


process_report/process_report.py
```diff
@@ -1,9 +1,108 @@
...
+        "uml.edu"               : "University of Massachusetts Lowell",
+        "codeforboston.org"     : "Code For Boston",
+        "mmsh"                  : "Harvard University",
+        "gstuart"               : "University of Massachusetts Amherst",
+        "rudolph"               : "Boston Childrens Hospital",
+        "robbaron"              : "Boston University",
+        "kmdalton"              : "Harvard University",
+        "mzink"                 : "University of Massachusetts Amherst",
+        "yale.edu"              : "Yale University",
+        "francesco.pontiggia"   : "Harvard University",
```
naved001, line 96: This and a bunch of others above it are not institute domains but rather people's name. Why do we have this here?

process_report/process_report.py
```diff
@@ -1,9 +1,108 @@
...
+    }
+
+    for name, institute in institute_map.items():
+        if name in pi_uname: return institute
```
naved001, line 100: what's an example of a `pi_uname`? curious how it looks like.

process_report/process_report.py
```diff
@@ -1,9 +1,108 @@
...
+    credit_amount = 1000
+
+    old_pi_list = set()
+    try:
+        with open(os.getenv("C0002_OLD_PI")) as f:
+            for pi in f: old_pi_list.add(pi.strip())
+
+    except Exception:
+        print("Old PI file does not exist, or the C0002_OLD_PI env var is not set")
+        sys.exit(1)
```
naved001, line 40: Couple of things about this block:
1. Check that "C0002_OLD_PI env var" is set earlier in the code, before trying to open the file. If it's not set, then raise an error and exit.
2. Then, when you open the file, then specifically check for the exception that the file doesn't exist (`FileNotFoundError`). Don't have an all encompassing exception check.

This way you are clear about what caused the failure, instead of saying it could be "X or Y" that caused the failure. For anything else, the exception will be printed.

process_report/process_report.py
```diff
@@ -1,9 +1,108 @@
...
+
+    pi_list = dataframe[MANAGER_PI_FIELD].unique()
+
+    for pi in pi_list:
+        if pi != pi: continue   # NaN check
```
naved001, line 45: can you say more about this check? The comment is not helpful.

naved001: Also, it's not a popular to write if/esle statements in one line. Please stick to the standard formatting. You could use [`black`](https://github.com/psf/black) to format your code; although it formats the whole file, so don't use it in this PR as it'll make the diff unreadable. Kust manually fix the if/else statements please.

process_report/process_report.py
```diff
@@ -1,9 +1,108 @@
...
+BALANCE_FIELD = 'Balance'
+###
+
+
+def apply_credits_0001(dataframe):
+    credit_code = "0001"
+    pass
+
+
+def apply_credits_0002(dataframe):
```
joachimweyl, line 27: isn't this already done in pull #12?

process_report/process_report.py
```diff
@@ -112,21 +228,80 @@ def remove_billables(dataframe, pi, projects, output_file):
...
+
+
+def export_lenovo(dataframe: pandas.DataFrame, output_file=None):
+
+    invoice_date = dataframe[INVOICE_MONTH_FIELD].iat[0].split('-')
+    invoice_year = invoice_date[0]
+    invoice_month = invoice_date[1]
+    lenovo_file_name = output_file or f'Lenovo_{invoice_year}_{invoice_month}.csv'
+
+    LENOVO_SU_TYPES = ['OpenShift GPUA100SXM4', 'OpenStack GPUA100SXM4']
```
joachimweyl, line 290: @knikolla are we flagging rows that would be useful to add to [this](https://github.com/nerc-project/operations/issues/420) repo once it is ready?

process_report/process_report.py
```diff
@@ -1,9 +1,108 @@
...
+    
+    return dataframe
+
+
+applied_credits = [apply_credits_0002]
+
+
+def get_institution_from_pi(pi_uname):
+    
+    institute_map = {
```
joachimweyl, line 71: @knikolla are we flagging rows that would be useful to add to [this](https://github.com/nerc-project/operations/issues/420) repo once it is ready?

process_report/tests/unit_tests.py
```diff
@@ -175,3 +176,101 @@ def test_export_pi(self):
...
+
+class TestExportLenovo(TestCase):
+    def setUp(self):
+
+        data = {
+            'Invoice Month': ['2023-01','2023-01','2023-01','2023-01','2023-01', '2023-01'],
+            'Project - Allocation': ['ProjectA', 'ProjectB', 'ProjectC', 'ProjectD', 'ProjectE', 'ProjectF'],
+            'Institution': ['A', 'B', 'C', 'D', 'E', 'F'],
+            'SU Hours (GBhr or SUhr)': [1, 10, 100, 4, 432, 10],
+            'SU Type': ['OpenShift GPUA100SXM4', 'CPU', 'OpenShift GPUA100SXM4', 'OpenStack GPUA100SXM4', 'CPU', 'GPU']
```
joachimweyl, line 249: instead of GPU can we put GPUA100, and instead of a second CPU can we put GPUV100

process_report/process_report.py
```diff
@@ -112,21 +228,80 @@ def remove_billables(dataframe, pi, projects, output_file):
...
+
+def export_lenovo(dataframe: pandas.DataFrame, output_file=None):
+
+    invoice_date = dataframe[INVOICE_MONTH_FIELD].iat[0].split('-')
+    invoice_year = invoice_date[0]
+    invoice_month = invoice_date[1]
+    lenovo_file_name = output_file or f'Lenovo_{invoice_year}_{invoice_month}.csv'
+
+    LENOVO_SU_TYPES = ['OpenShift GPUA100SXM4', 'OpenStack GPUA100SXM4']
+    SU_CHARGE_MULTIPLIER = 1
```
joachimweyl, line 291: @knikolla are we flagging rows that would be useful to add to [this](https://github.com/nerc-project/operations/issues/420) repo once it is ready?

process_report/process_report.py
```diff
@@ -112,21 +228,80 @@ def remove_billables(dataframe, pi, projects, output_file):
...
+                               (dataframe[INSTITUTION_FIELD] == 'Boston University')]
+    HU_BU_projects.to_csv(output_file)        
+
+
+def export_lenovo(dataframe: pandas.DataFrame, output_file=None):
+
+    invoice_date = dataframe[INVOICE_MONTH_FIELD].iat[0].split('-')
+    invoice_year = invoice_date[0]
+    invoice_month = invoice_date[1]
+    lenovo_file_name = output_file or f'Lenovo_{invoice_year}_{invoice_month}.csv'
```
naved001, line 288: one liner could be:
```python
year, month = dataframe[INVOICE_MONTH_FIELD].iat[0].split('-')
```
Though, do we really need to do this just so we can go from `Lenovo_year_month.csv` to `Lenovo_year-month.csv` ?

naved001 (changes requested):

process_report/process_report.py
QuanMPhm: @naved001 While running tests on past invoice CSVs, I found that some project allocations did not have a value in their PI field. When `pandas` read the invoice CSV, it assigns those empty fields with the value `NaN`. Using `if pi != pi` was one way I found to check whether the PI field was empty or not. I've updated my PR in #12 to use `pandas.isna()`, which should make the intention more clear.

process_report/process_report.py
QuanMPhm: @naved001 Some PIs have set their usernames to be...usernames, not they institute-specific email. As far as I know, the only way to know what institute they belong to is to hardcode their usernames in this mapping.

process_report/process_report.py
naved001: It looks like the order of the the dictionary matters in the way this loop works.

For instance, if we have a PI `somebody@dfci.harvard.edu` then `dfci.harvard.edu` must be checked before `harvard.edu` because `name in pi_uname` is true for both cases.  So whoever is modifying the institute mapping file will have to be quite careful to not undo this order. 

I think this would be a faster (no loop) and safer (looking for exact match) way to do this: 

```python3
def get_institution_from_pi(pi_uname, institute_map):
if '@' in pi_uname:
    domain = pi_uname.split('@')[1]
    return institute_map.get(domain, '')

return institute_map.get(pi_uname, '')
```

What do you think?

process_report/process_report.py
QuanMPhm: @naved001 I've turned the naming into a one-liner

QuanMPhm:
@naved001 @joachimweyl  I've addressed all of comments so far, except for ones involving @knikolla

process_report/process_report.py
knikolla: @joachimweyl good suggestion. Not yet. For now tracking it here seems fine.

process_report/tests/unit_tests.py
QuanMPhm: @joachimweyl  For the purpose of this test, I don't think replacing the string `GPU` with `GPUA100` would make a difference, since to know if a project used a Lenovo SU type, the code tries to match the SU Type exactly to either `OpenShift GPUA100SXM4` or `OpenStack GPUA100SXM4`

joachimweyl: It is not supposed to, but the point of a test is to test all possibilities right?
To be clear we should put `OpenShift GPUA100` and `OpenStack GPUV100`. and `CPU` should actually be `OpenShift CPU` or `OpenStack CPU`.  We also might want to test that it properly combines `OpenStack GPUA100SXM4` so maybe replace one of the other ones with that instead.

process_report/process_report.py
QuanMPhm: @naved001 After applying your suggestion and running tests on my script again... By looking for exact matches instead of substrings, we might miss cases where the PIs email domain might be a subdomain of an institution (i.e `fas.harvard.edu` is still Harvard University, not an affiliate). This would mean we would have to add more cases for subdomains of the same institution. Do you think it's still worth it to do exact matches.

naved001: We already have a bunch of subdomains in the institution_map, I think it's okay to update the institution_map with new subdomains when we encounter those.

I thought maybe we could extract the top level domain from the string, but apparently that requires querying all TLDs  which I think is not worth it in our case. See: https://github.com/john-kurkowski/tldextract?tab=readme-ov-file

QuanMPhm: I've applied your suggestion in the [other](https://github.com/CCI-MOC/process_csv_report/pull/12) PR

process_report/tests/unit_tests.py
QuanMPhm: @joachimweyl What do you mean by properly combines `OpenStack GPUA100SXM4`?

joachimweyl: "calculates" not combines, sorry.

QuanMPhm: I've updated the list of SU Types being tested.

naved001:
@QuanMPhm once #12 merges, please rebase this PR

naved001 (approved):

knikolla (approved):

PR #20, opened 2024-04-18
Added a script to clone nonbillable information using deploy keys

QuanMPhm:
The script requires the environment variable `GH_NONBILLABLE_DEPLOYKEY` to be set, containg the private key of the deploy key pair Execution of the script will clone the nonbillable repo, and copy the relevant .txt files into the `process_csv_report` repo


clone_nonbillables_and_process.sh
```diff
@@ -0,0 +1,19 @@
...
+# Add deploy key to ssh config
+mkdir -p ~/.ssh
+touch ~/.ssh/config
+touch ~/.ssh/id_nonbillable
+if ! grep -q "Host github-nonbillable" ~/.ssh/id_nonbillable; then
+    echo "
+    Host github-nonbillable
+        HostName github.com
+        IdentityFile ~/.ssh/id_nonbillable
+    " > ~/.ssh/config
```
knikolla, line 13: If this is run from someone's machine, rather than a temporary container, it is going to destroy their `.ssh/config` file. Is there a way to do this without changing the ssh config file?

knikolla (changes requested):

clone_nonbillables_and_process.sh
naved001: good catch, one could do `>> ~/.ssh/config` to append to file instead.

knikolla: Actually, my preference would be to stop execution if it detects the presence of a `~/.ssh/config` file.

If such a file exists, then this is running on someone's machine rather than a temporary container. In which case they probably already have an SSH key that has access to that repository already configured.

clone_nonbillables_and_process.sh
```diff
@@ -0,0 +1,19 @@
...
+    echo "$GH_NONBILLABLE_DEPLOYKEY" > ~/.ssh/id_nonbillable
+fi
+chmod 600 ~/.ssh/id_nonbillable
+
+git clone git@github-nonbillable:CCI-MOC/non-billable-projects.git ~/non-billable-projects
+cp ~/non-billable-projects/*.txt .
```
knikolla, line 19: You shouldn't do this. Treat the repository as the single source of truth and pass-in the path to the files inside the folder when calling `process_csv_report`.

knikolla (changes requested):

knikolla:
Ah, and one final comment. Since there are likely going to be other scripts that we have to write. I would place this in a directory rather than directly in the root. Choose something like `tools`, `bin`.

clone_nonbillables_and_process.sh
QuanMPhm: I have removed the `cp`. As for passing the files into `process_report.py`, I'll be handling that part in #21.

clone_nonbillables_and_process.sh
QuanMPhm: I changed the if statement to check if `~/.ssh/config` exists instead.

tools/clone_nonbillables_and_process.sh
```diff
@@ -0,0 +1,18 @@
...
+    echo "
+    Host github-nonbillable
+        HostName github.com
+        IdentityFile ~/.ssh/id_nonbillable
+    " > ~/.ssh/config
+    echo "$GH_NONBILLABLE_DEPLOYKEY" > ~/.ssh/id_nonbillable
+fi
+chmod 600 ~/.ssh/id_nonbillable
+
+git clone git@github-nonbillable:CCI-MOC/non-billable-projects.git ~/non-billable-projects
```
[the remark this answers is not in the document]
knikolla: I would also advise you in general, to restrict yourself from creating files in someone's home folder. Do the cloning in this current directory where the script is executed from.

knikolla (approved):
Thanks! Looks good to me!

PR #26, opened 2024-04-23
Implemented processing and exporting for BU Internal invoice

### LINKED ISSUE #25: Create a new HU & BU only Invoice & Create a BU Internal Invoice CSV
## Motivation
Automate the creation of a CSV of the Invoice sheet that the BU billing team requires for creating internal invoicing. Also, create a CSV of only the institutions that are currently being billed (HU & BU).

## Completion Criteria
CSV output of the BUInternalInvoice sheet created at the same time as the invoice scripts. CSV output of only the currently being billed Invoice data, currently only BU & HU.

### BU & HU CSV
- [x] Create a CSV that filters on only BU & HU Addresses
- [ ] test that it matches the data in the original Invoice CSV

### BU Internal Invoice
- [x] Review the [BUInternalInvoice sheet](https://docs.google.com/spreadsheets/d/1ZZELpyIpmcmCCwZXt3wiccKke8ljyW_WFOyd1NnMIVY/edit?pli=1#gid=1763617817)
- [x] Column 1 is Invoice Month
- [x] Column 2 is Manager PI
- [x] Column 3 is Project, can pull from ColdFront or filter Project - Allocation to remove the Allocation, in google sheets that can be done with this code `=REGEXEXTRACT(D2, "(.+)-")`
- [ ] then remove duplicate Project rows
- [ ] Column 4 is Cost which is a sum of all Cost from the main CSV where Project - Allocation includes the string that is in column 3 or from coldfront just the rows that match that Project
- [x] Column 5 is Credit same as column 4 but for Credit
- [x] Column 6 is Subsidy this one is a bit complex you can look at the function in the sheet or you can follow these rules
   - Each PI receives up to $100 discount across all their projects
   - If they have less than $100 usage then whatever they used is subsidized
   - If they used more than $100 remove up to $100 from each row of their projects until $100 is removed
- [x] Column 7 is Balance = Cost - Credit - Subsidy

## Completion dates
Desired - 2024-04-25
Required - TBD

QuanMPhm:
Closes #25 fully, I have implemented the function to export the BU Internal invoice, following @joachimweyl's specifications in the linked issue.

The BU subsidy will be applied to every BU PI after the New-PI credit is applied to eligible PIs in the combined invoice. With how this is currently implemented, the combined invoice may NOT show BU PIs' final balances, as the BU-only invoice is made and exported after the combined invoice is exported.

The current subsidy is set at $100, but I have made the amount configurable via a cli arguement.

### DIFF
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -24,6 +24,7 @@ SU_TYPE_FIELD = "SU Type"
 COST_FIELD = "Cost"
 CREDIT_FIELD = "Credit"
 CREDIT_CODE_FIELD = "Credit Code"
+SUBSIDY_FIELD = "Subsidy"
 BALANCE_FIELD = "Balance"
 ###
 
@@ -163,6 +164,12 @@ def main():
         help="Name of output folder containing pi-specific invoice csvs",
     )
     parser.add_argument(
+        "--BU-invoice-file",
+        required=False,
+        default="BU_Internal.csv",
+        help="Name of output csv for BU invoices",
+    )
+    parser.add_argument(
         "--HU-BU-invoice-file",
         required=False,
         default="HU_BU.csv",
@@ -178,6 +185,12 @@ def main():
         "--old-pi-file",
         required=False,
         help="Name of csv file listing previously billed PIs. If not provided, defaults to fetching from S3",
+    )
+    parser.add_argument(
+        "--BU-subsidy-amount",
+        required=True,
+        type=int,
+        help="Amount of subsidy given to BU PIs",
     )
     args = parser.parse_args()
 
@@ -218,6 +231,7 @@ def main():
 
     export_billables(credited_projects, args.output_file)
     export_pi_billables(credited_projects, args.output_folder, invoice_month)
+    export_BU_only(billable_projects, args.BU_invoice_file, args.BU_subsidy_amount)
     export_HU_BU(credited_projects, args.HU_BU_invoice_file)
     export_lenovo(credited_projects, args.Lenovo_file)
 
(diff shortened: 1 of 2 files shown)


process_report/process_report.py
```diff
@@ -302,6 +310,55 @@ def export_HU_only(dataframe, output_file):
...
+            INVOICE_DATE_FIELD,
+            PI_FIELD,
+            "Project",
+            COST_FIELD,
+            CREDIT_FIELD,
+            SUBSIDY_FIELD,
+            BALANCE_FIELD,
+        ]
+    ]
+    BU_projects = _apply_BU_subsidy(BU_projects)
```
knikolla, line 335: Running this here (after already having exported the billable file) is going to cause the balance fields to be different in the billables invoice and the BU only invoice. @joachimweyl is this something we care about?

process_report/process_report.py
```diff
@@ -105,6 +106,12 @@ def main():
         default="HU_only.csv",
         help="Name of output csv for HU invoices",
     )
+    parser.add_argument(
+        "--BU-invoice-file",
+        required=False,
+        default="BU_Internal.csv",
+        help="Name of output csv for HU invoices",
```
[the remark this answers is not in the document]
QuanMPhm: I (obliviously) hope this is the only feedback you have?

process_report/process_report.py
```diff
@@ -302,6 +310,55 @@ def export_HU_only(dataframe, output_file):
...
+            SUBSIDY_FIELD,
+            BALANCE_FIELD,
+        ]
+    ]
+    BU_projects = _apply_BU_subsidy(BU_projects)
+    BU_projects.to_csv(output_file)
+
+
+def _apply_BU_subsidy(dataframe):
+    subsidy_amount = 100
```
naved001, line 340: @joachimweyl could the subsidy amount change in future? @QuanMPhm If it could change maybe this should be an argument that's passed down?

process_report/process_report.py
```diff
@@ -302,6 +310,55 @@ def export_HU_only(dataframe, output_file):
...
+            CREDIT_FIELD,
+            SUBSIDY_FIELD,
+            BALANCE_FIELD,
+        ]
+    ]
+    BU_projects = _apply_BU_subsidy(BU_projects)
+    BU_projects.to_csv(output_file)
+
+
+def _apply_BU_subsidy(dataframe):
```
naved001, line 339: there's nothing BU specific in this method as far as I can tell, looks like it'll apply to any projects that are in the dataframe. So, just generalize it and we could re-use this method if we were to apply subsidies to any other institutions?

naved001 (commented):
@QuanMPhm could you update the commit message with a brief description of how this is supposed to work? For example, I don't remember if subsidies apply before credits or the other way around etc.

process_report/process_report.py
QuanMPhm: I've added a cli argument for the user to specify the subsidy amount

knikolla: While the subsidy function is pretty general and it made sense to rename, this specific instance of the subsidy passed as argument is only applied to BU, so the CLI argument name for this specific subsidy should reflect that.

process_report/process_report.py
```diff
@@ -165,6 +172,11 @@ def main():
         required=False,
         help="Name of csv file listing previously billed PIs",
     )
+    parser.add_argument(
```
knikolla, line 175: You should specify a type here as int, this way you don't need to typecast it later in the program.

process_report/process_report.py
```diff
@@ -386,6 +399,54 @@ def export_HU_only(dataframe, output_file):
...
+def _apply_subsidy(dataframe, subsidy_amount):
+    pi_list = dataframe[PI_FIELD].unique()
+
+    for pi in pi_list:
+        pi_projects = dataframe[dataframe[PI_FIELD] == pi]
+        remaining_subsidy = subsidy_amount
+        for i, row in pi_projects.iterrows():
+            project_remaining_cost = row[
+                BALANCE_FIELD
+            ]  # Applying BU Subsidy after applying credits, so Balance shows rem costs
```
knikolla, line 437: This function doesn't know (nor should it know) when it's being called therefore this comment is wrong. The BU Subsidy is applied after credits *because* of when the `export_BU_only` is called.

knikolla (changes requested):

knikolla (approved):
Looks fine to me. @naved001 can you do a final pass?

process_report/tests/unit_tests.py
```diff
@@ -329,6 +329,83 @@ def test_apply_credit_0002(self):
...
+
+        self.assertTrue(
+            set(["PI1", "PI2", "PI4"]).issubset(output_df["Manager (PI)"].unique())
+        )
+
+        self.assertTrue(
+            set(["ProjectA", "ProjectB", "ProjectE-test", "ProjectF"]).issubset(
+                output_df["Project"].unique()
+            )
+        )
```
naved001, line 395: It would be nice to add assertions that PI3 isn't in the output and their projects `ProjectC-t14334` and `ProjectD` are also not in the output just to make the test complete. Because if the harvard PIs and their projects were in the output these assertions will still be good.

```
>>> projects = ["a", "b", "c", "d", "e"]
>>> set(["a", "b", "e"]).issubset(projects)
True
```

naved001 (commented):
just one suggestion for improving the tests, other than that it looks good to me.

naved001 (approved):

Issue #21, opened 2024-04-18
Create Dockerfile to containerize billing script


QuanMPhm:
Splitting up #15, and dependent on #20, as the next step before creating a fully automated cronjob to process invoices, we want to create a Docker container containing the bare dependencies needed to run the cloning and billing script.

QuanMPhm:
@knikolla @naved001 As a solution I will create a Docker container with a shell script which does the following:
- Clone the non-billable repo
- Fetches the invoices for the previous month from S3 storage
- Passes all relevant files to `process_report.py`

knikolla:
> @knikolla @naved001 As a solution I will create a Docker container with a shell script which does the following:
> 
> * Clone the non-billable repo
> * Fetches the invoices for the previous month from S3 storage
> * Passes all relevant files to `process_report.py`

@QuanMPhm And 4th step, uploads the generated invoices in the appropriately named way.

QuanMPhm:
My questions/things I need so far are:
- I need access to a list of Old PIs. Where is that currently stored? @naved001 suggested that it should be placed in the nonbillable repo alongside other files that would impact billing.
- @joachimweyl @knikolla  Since we want this Docker container to upload the processed invoices back to S3, what should be the naming convention for all the HU-only, HU & BU only, Lenovo-only, and the PI-specific invoices?

QuanMPhm:
@joachimweyl If it's not a big problem, I'll have the file names be as follows:
```
HU&BU - NERC Total Invoice <YYYY>-<MM>.csv
BU Internal - NERC BU <YYYY>-<MM>.csv
Lenovo-only - Lenovo <YYYY>-<MM>.csv
PI specific - PI Invoices/<University>_<PI>_<YYYY>-<MM>.csv
```

It would juts simplify the code if the invoice month (YYYY-MM) is consistently placed at the end of the filename.

joachimweyl:
Please make sure HU&BU is labeled `NERC-<YYYY>-<MM>-Total-Invoice.csv` as that is the file name MGHPCC requested. The others look fine.

PR #28, opened 2024-04-24
Added option to fetch and upload old PI file to from and to S3, and added logic to update old PI file

### LINKED ISSUE #27: Fetch and update old pi file to s3
Required by #21 and #22 

- Need to generate an initial old pi file containing pi names and month.
- Upload this initial pi file to B2 in a known location manually.
- Update the code to fetch the old pi file.
- Update the code to write a new pi file with the new pis encountered this month.
- Upload the old (new) pi file to s3.

QuanMPhm:
Partially closes #27. If the user does not provide an old_pi file, `process_report.py` will by default fetch the file from a hardcoded location in S3 storage.
Setting the `upload-to-s3` option will now upload both the processed CSV invoices and the updated PI file.

I've also renamed some functions for consistent capitalization.

I have also added a few print statements to provide the following info:
- When a new PI is encountered in an invoice
- When a future PI is found in a past invoice

It should be noted that if future PIs are encountered when processing old invoices, the invoice script will consider them new PIs for the purpose of the new PI credit, and will update the future PI's first-month date to be same as the old invoice's date.

Blocked on #23, and prerequisite to #21

### DIFF
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -60,10 +60,28 @@ def load_old_pis(old_pi_file):
     return old_pi_dict
 
 
+def dump_old_pis(old_pi_file, old_pi_dict: dict):
+    with open(old_pi_file, "w") as f:
+        for pi, first_month in old_pi_dict.items():
+            f.write(f"{pi},{first_month}\n")
+
+
 def is_old_pi(old_pi_dict, pi, invoice_month):
-    if pi in old_pi_dict and old_pi_dict[pi] != invoice_month:
+    first_invoice_month = old_pi_dict.get(pi, invoice_month)
+    if compare_invoice_month(first_invoice_month, invoice_month):
+        sys.exit(
+            f"PI {pi} from {first_invoice_month} found in {invoice_month} invoice!"
+        )
+    if compare_invoice_month(invoice_month, first_invoice_month):
         return True
     return False
+
+
+def compare_invoice_month(month_1, month_2):
+    """Returns True if 1st date is later than 2nd date"""
+    dt1 = datetime.datetime.strptime(month_1, "%Y-%m")
+    dt2 = datetime.datetime.strptime(month_2, "%Y-%m")
+    return dt1 > dt2
 
 
 def get_invoice_bucket():
@@ -103,7 +121,7 @@ def main():
     parser.add_argument(
         "--upload-to-s3",
         action="store_true",
-        help="If set, uploads all processed invoices to S3",
+        help="If set, uploads all processed invoices and old PI file to S3",
     )
     parser.add_argument(
         "--invoice-month",
@@ -159,16 +177,20 @@ def main():
     parser.add_argument(
         "--old-pi-file",
         required=False,
-        help="Name of csv file listing previously billed PIs",
+        help="Name of csv file listing previously billed PIs. If not provided, defaults to fetching from S3",
     )
     args = parser.parse_args()
 
     invoice_month = args.invoice_month
 
     if args.fetch_from_s3:
-        csv_files = fetch_S3_invoices(invoice_month)
+        csv_files = fetch_s3_invoices(invoice_month)
     else:
         csv_files = args.csv_files
+    if args.old_pi_file:
+        old_pi_file = args.old_pi_file
+    else:
+        old_pi_file = fetch_s3_old_pi_file()
 
     merged_dataframe = merge_csv(csv_files)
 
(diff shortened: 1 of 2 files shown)


process_report/process_report.py
```diff
@@ -58,7 +58,14 @@ def load_old_pis(old_pi_file):
...
 
 
+def dump_old_pis(old_pi_file, old_pi_dict: dict):
+    with open(old_pi_file, "w") as f:
+        for pi, first_month in old_pi_dict.items():
+            f.write(f"{pi},{first_month}\n")
+
+
 def is_old_pi(old_pi_dict, pi, invoice_month):
+    # If we're processing old invoices, this will flag PIs who were new *back then* as old. But this would only be the case if future PIs appeared in past invoices???
```
knikolla, line 68: - Do not write lines this long, break the comment into multiple line comments.
- Comments are not a place to have a dialogue with yourself. An issue with the code should be flagged as an issue, or resolved.
- You can easily turn this check from a `!=` to comparing if the date is in the past or future. 
  - Write a unit test for it.

process_report/process_report.py
```diff
@@ -217,9 +228,10 @@ def main():
             invoice_list.append(os.path.join(args.output_folder, pi_invoice))
 
         upload_to_s3(invoice_list, invoice_month)
+        upload_to_s3_old_pi_file(old_pi_file)
 
 
-def fetch_S3_invoices(invoice_month):
+def fetch_s3_invoices(invoice_month):
```
knikolla, line 234: While yes, this should have been `s3` instead of `S3` and in python things are preferred to be written in lower case snake case. Please try to refrain from introducing unrelated changes to a PR. Do not revert it for now, this is fine.

knikolla (changes requested):

QuanMPhm:
I've added some small features and have updated the PR summary

process_report/process_report.py
```diff
@@ -58,8 +58,16 @@ def load_old_pis(old_pi_file):
...
+def dump_old_pis(old_pi_file, old_pi_dict: dict):
+    with open(old_pi_file, "w") as f:
+        for pi, first_month in old_pi_dict.items():
+            f.write(f"{pi},{first_month}\n")
+
+
 def is_old_pi(old_pi_dict, pi, invoice_month):
-    if pi in old_pi_dict and old_pi_dict[pi] != invoice_month:
+    if pi in old_pi_dict and old_pi_dict[pi] > invoice_month:
+        print(f"Found future PI {pi} from {old_pi_dict[pi]} in {invoice_month} invoice")
```
knikolla, line 69: This print is not necessary and this function can be simplified into only one `if` statement by using `.get` on the dictionary and providing a default value of the current month.

knikolla (changes requested):

process_report/process_report.py
QuanMPhm: We don't need to know if a future PI is encountered?

knikolla: There's no such thing as a "future" PI. If when re-processing the invoices of the past, you encounter a PI whose first encountered invoice is later than the invoice you're processing, there's something wrong with the invoices or the PI file. If you want to be extra careful you can raise an exception here and just stop the script altogether with an error.

QuanMPhm: I've made it raise an exception, and added a test case for it.

process_report/process_report.py
```diff
@@ -58,8 +58,16 @@ def load_old_pis(old_pi_file):
...
 
+def dump_old_pis(old_pi_file, old_pi_dict: dict):
+    with open(old_pi_file, "w") as f:
+        for pi, first_month in old_pi_dict.items():
+            f.write(f"{pi},{first_month}\n")
+
+
 def is_old_pi(old_pi_dict, pi, invoice_month):
-    if pi in old_pi_dict and old_pi_dict[pi] != invoice_month:
+    if pi in old_pi_dict and old_pi_dict[pi] > invoice_month:
```
knikolla, line 68: Rewrite this block as below to improve readability. 
```python
first_invoice_month = old_pi_dict.get(pi, invoice_month)
if first_invoice_month > invoice_month:
    ... exit with the error here
if first_invoice_month < invoice_month:
    return True
return False
```


```

knikolla (changes requested):
One last comment, otherwise looks good.

knikolla (approved):
Looks good to me. @naved001 can you do a final pass today?

naved001 (changes requested):

QuanMPhm:
@knikolla @naved001 Which PR should be pushed first?

naved001:
@QuanMPhm maybe the order you proposed the PRs. You'll need to fix the merge conflicts on both PRs though.

PR #30, opened 2024-04-25
Removed HU invoice processing and added S3 export function for HU-BU invoice

### LINKED ISSUE #29: Remove HU-only invoice processing, and add s3 upload function to specifically upload HU-BU invoice
Based on @joachimweyl's comments [here](https://github.com/CCI-MOC/process_csv_report/issues/21#issuecomment-2075573646) and [here](https://github.com/CCI-MOC/process_csv_report/issues/21#issuecomment-2069812912), we no longer need to export a HU-only invoice, and will need to add a function to specifically export the HU-BU invoice. A dedicated function is needed because the HU-BU invoice filename's pattern is different from all other invoices being uploaded to S3, due to the MGHPCC's request.

QuanMPhm:
Closes #29, I've removed HU invoice processing, and added an S3 export function to specifically export the HU-BU invoice. Since the HU-BU invoice will now be exported to a (mostly) hardcoded path in S3, the user's choice of filename for the HU-BU invoice will no longer matter when it is exported to S3.

### DIFF
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -143,12 +143,6 @@ def main():
         help="Name of output folder containing pi-specific invoice csvs",
     )
     parser.add_argument(
-        "--HU-invoice-file",
-        required=False,
-        default="HU_only.csv",
-        help="Name of output csv for HU invoice",
-    )
-    parser.add_argument(
         "--HU-BU-invoice-file",
         required=False,
         default="HU_BU.csv",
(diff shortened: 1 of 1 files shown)


knikolla:
@QuanMPhm The button to ask for reviews is different from the button to "assign" a PR.

knikolla (approved):

naved001 (approved):

PR #31, opened 2024-04-25
Containerized billing script

### LINKED ISSUE #21: Create Dockerfile to containerize billing script

QuanMPhm:
Closes #21 completely. This PR is blocked on #27, and will containerize the billing script. A github workflow is added to build and publish the billing image. Check the second commit to see the changes which actually implemented the containerization.

### DIFF
--- a/tools/clone_nonbillables_and_process.sh
+++ b/tools/clone_nonbillables_and_process.sh
@@ -29,8 +29,14 @@ fi
 
 INVOICE_MONTH=$(date --date="$(date +%Y-%m-01) -1 month" +%Y-%m)
 python process_report/process_report.py \
+    --fetch-from-s3 \
+    --upload-to-s3 \
     --invoice-month $INVOICE_MONTH \
     --pi-file ./non-billable-projects/pi.txt \
     --projects-file ./non-billable-projects/projects.txt \
     --timed-projects-file ./non-billable-projects/timed_projects.txt \
-    --old-pi-file old_pi.csv
+    --nonbillable-file "NERC (Nonbillable).csv" \
+    --output-file "NERC.csv" \
+    --output-folder "PI Invoices" \
+    --BU-invoice-file "NERC BU.csv" \
+    --Lenovo-file "Lenovo.csv"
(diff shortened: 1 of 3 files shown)


tools/clone_nonbillables_and_process.sh
```diff
@@ -28,9 +28,19 @@ if [ ! -d ./non-billable-projects ]; then
 fi
 
 INVOICE_MONTH=$(date --date="$(date +%Y-%m-01) -1 month" +%Y-%m)
+export S3_ENDPOINT
```
knikolla, line 31: These exports are unnecessary as these valued will not be assigned in the script but outside of it.

tools/clone_nonbillables_and_process.sh
```diff
@@ -28,9 +28,19 @@ if [ ! -d ./non-billable-projects ]; then
...
+    --upload-to-s3 \
     --invoice-month $INVOICE_MONTH \
     --pi-file ./non-billable-projects/pi.txt \
     --projects-file ./non-billable-projects/projects.txt \
     --timed-projects-file ./non-billable-projects/timed_projects.txt \
-    --old-pi-file old_pi.csv
+    --nonbillable-file "NERC (Nonbillable)" \
+    --output-file "NERC" \
+    --output-folder "PI Invoices" \
+    --BU-invoice-file "NERC BU" \
```
knikolla, line 45: You are missing the file extensions in all arguments that are a file here.

knikolla (changes requested):

tools/clone_nonbillables_and_process.sh
QuanMPhm: Since all the files in the container will be ephemeral, why do we need to include extensions?

knikolla: You are assuming that this script will only ever be used inside the container.

naved001 (approved):

PR #33, opened 2024-04-29
Packed invoice container into Kubernetes cronjob

### LINKED ISSUE #22: Package the invoice processing container into a Openshift Cronjob
Dependant on #21, the goal is to create a Cronjob that runs at the beginning of each month which performs the following tasks in order:
- Fetch monthly usage CSVs from S3
- Fetch nonbillable data from the CCI-MOC private repo
- Fetch the list of old PIs from S3 (maybe this file should be relocated to our nonbillable Github repo?)
- Process the invoices
- Upload the processed to S3

I have  a question regarding the precise details, and I'll likely add more as I progress through this issue:
- @joachimweyl Currently the invoice processing script exports not only the combined Billable and Non-Billable invoices, but also the HU-only, HU & BU only, Lenovo-only, and the PI-specific invoices as well. Should we decide on a naming convention for these files and how they should be stored in S3?

QuanMPhm:
Blocked on #31. Closes #22. The manifests needed for the cronjob includes the cronjob, and two secret files which sets the private key for the nonbillable repo's deploykey and the s3 app_key and key_id.

The cronjob is set to run at 12:00PM (Noon) on the 1st of each month. The last commit will show what's actually included in this PR.

### DIFF
--- a/k8/base/invoice-processing.yaml
+++ b/k8/base/invoice-processing.yaml
@@ -1,0 +1,30 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: invoice-processing
+spec:
+  schedule: "0 14 1 * *"
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+          - name: invoice-processing
+            image: ghcr.io/cci-moc/process-csv-report:latest
+            env:
+              - name: GH_NONBILLABLE_DEPLOYKEY
+                valueFrom:
+                secretKeyRef:
+                  name: gh-nonbillables
+                  key: ssh-deploykey
+              - name: S3_KEY_ID
+                valueFrom:
+                secretKeyRef:
+                  name: nerc-invoices-s3-bucket
+                  key: s3-key-id
+              - name: S3_APP_KEY
+                valueFrom:
+                secretKeyRef:
+                  name: nerc-invoices-s3-bucket
+                  key: s3-app-key
+          restartPolicy: OnFailure
--- a/k8/base/kustomization.yaml
+++ b/k8/base/kustomization.yaml
@@ -1,0 +1,2 @@
+resources:
+  - invoice-processing.yaml
--- a/k8/overlay/gh-nonbillables.yaml
+++ b/k8/overlay/gh-nonbillables.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: gh-nonbillables
+type: kubernetes.io/ssh-auth
+data:
+  ssh-deploykey: test
--- a/k8/overlay/kustomization.yaml
+++ b/k8/overlay/kustomization.yaml
@@ -1,0 +1,4 @@
+resources:
+  - ../base
+  - secret-b2-old-pi.yaml
+  - secret-gh-nonbillables.yaml
--- a/k8/overlay/nerc-invoices-s3-bucket.yaml
+++ b/k8/overlay/nerc-invoices-s3-bucket.yaml
@@ -1,0 +1,8 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: nerc-invoices-s3-bucket
+type: Opaque
+data:
+  s3-app-key: test
+  s3-key-id: test


k8/base/cronjob.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: hello
+spec:
+  schedule: "0 12 1 * *"
```
knikolla, line 6: This should be something like 8.30am rather than midnight to give time to the other invoicing scripts to finish their job and place everything in S3.

QuanMPhm: Sorry, I made a typo, forgetting that "12AM meant midnight. " `0 12 1 * *` would mean running at Noon, `0 0 1 * *` would have been midnight

naved001: I run my cronbjo on the nerc cluster and those are all doing things in UTC. So, this would be 12noon at UTC or 8am boston time if run on a nerc openshift cluster.

FWIW, my cronjob produces the openshift report at 9 am Boston time. I am happy to change that if you want, but this should run after the openshift, openstack and coldfront reports are uploaded. 

I'd vote to change it to produce the final report at 10am Boston time or 2pm UTC, so `0 14 1 * *`.

QuanMPhm: Should I specify a Boston timezone in the cronjob manifest?

k8/base/cronjob.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: hello
```
larsks, line 4: Please pick a name that identifies the purpose of the cron job (otherwise someone will see resource named "hello" and wonder what is going on).

.github/workflows/build.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      # Login against a Docker registry except on PR
+      # https://github.com/docker/login-action
+      - name: Log into registry ${{ env.REGISTRY }}
+        if: github.event_name != 'pull_request'
+        uses: docker/login-action@v2
```
larsks, line 39: This should now be using `v3` of the `docker/login-action` action:

```suggestion
        uses: docker/login-action@v3
```

.github/workflows/build.yaml
```diff
@@ -0,0 +1,68 @@
...
+  build:
+
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
```
larsks, line 33: This should now be using v4 of the `actions/checkout` action:

```suggestion
        uses: actions/checkout@v3
```

.github/workflows/build.yaml
```diff
@@ -0,0 +1,68 @@
...
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      # Extract metadata (tags, labels) for Docker
+      # https://github.com/docker/metadata-action
+      - name: Extract Docker metadata
+        id: meta
+        uses: docker/metadata-action@v4
```
larsks, line 49: This should now be using v5 of the `docker/metadata-action` action:

```suggestion
        uses: docker/metadata-action@v5
```

.github/workflows/build.yaml
```diff
@@ -0,0 +1,68 @@
...
+            type=semver,pattern=v{{major}}.{{minor}}
+            type=semver,pattern=v{{major}}
+            type=ref,event=branch
+            type=ref,event=pr
+            type=sha
+
+      # Build and push Docker image with Buildx (don't push on PR)
+      # https://github.com/docker/build-push-action
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v4
```
larsks, line 63: This should now be using v5 of the `docker/build-push-action` action:

```suggestion
        uses: docker/build-push-action@v5
```

Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+WORKDIR /app
+
+RUN apt-get update && apt-get install -y git
+
+COPY tools/ tools/
+COPY requirements.txt .
+COPY process_report/process_report.py process_report/
+COPY process_report/institute_map.json process_report/
+
+RUN pip install -r requirements.txt
```
larsks, line 12: Always handle dependency installation first; this allows your container runtime to cache this step so that you don't need to re-install requirements every time you modify your code and rebuild the image.

```suggestion
COPY requirements.txt .
RUN pip install -r requirements.txt

COPY tools/ tools/
COPY process_report/process_report.py process_report/
COPY process_report/institute_map.json process_report/
```

k8/base/invoice-processing.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: cronjob
```
larsks, line 4: Please give this cronjob a meaningful name.

process_report/process_report.py
```diff
@@ -333,9 +346,23 @@ def apply_credits_new_pi(dataframe, old_pi_file):
...
+def fetch_s3_old_pi_file():
+    local_name = "PI.csv"
+    invoice_bucket = get_invoice_bucket()
+    invoice_bucket.download_file("PIs/PI.csv", local_name)
+    return local_name
+
+
+def upload_to_s3_old_pi_file(old_pi_file):
+    invoice_bucket = get_invoice_bucket()
+    invoice_bucket.upload_file(old_pi_file, "PIs/PI.csv")
```
larsks, line 363: You're using the literal value `"PIs/PI.csv"` in two different functions. This should be a constant/global instead.

QuanMPhm: For some of your suggested changes which are not directly related to the cronjob (i.e for this, and for the Dockerfile), they do relate to PRs that have closed previously. Should I open them and changes to those PRs, or what is the right way to handle this?

larsks: If you don't think it's appropriate to address a comment in this PR, then it's fine to fix them in a future PR. A common workflow is to open an issue with the comment, and include a link to the issue in a response to the comment (and then at some point in the future, open a pull request that closes that issue).

QuanMPhm: Should the fix to the action file, Dockerfile, and this line of code that you highlighted, be placed in seperate PRs, or would it be fine for them to be seperate commits bundled in one PR?

larsks: I think it's fine to roll up a collection of small maintenance changes into a single pull request. :+1: for using separate commits for unrelated changes.

naved001 (approved):

larsks (approved):

Issue #35, opened 2024-05-01
Standardize on precision of numbers for invoicing


knikolla:
Capture a decision on the amount of precision (digits after the decimal point) we want for the below values

- Rate
- Cost
- Balance

joachimweyl:
Rate - 3 decimal places
Cost - 2 decimal places
Balance - 2 decimal places

naved001:
I have been rounding the cost to 2 decimal places.

https://github.com/CCI-MOC/openshift-usage-scripts/blob/be997dc7f2332a6b04e08c03f595abba1912710b/openshift_metrics/utils.py#L368


Specifically, I have been using ROUND_HALF_UP because that's what I learnt in maths in school.

knikolla:
@joachimweyl For storage, the CSV processes per GB rather per TB, 3 decimal points for the rate aren't enough for that. I can [fix the presentation](https://github.com/nerc-project/coldfront-plugin-cloud/issues/157).

joachimweyl:
sorry you are completely correct let me fix my comment.

joachimweyl:
@naved001 what is ROUND_HALF_UP is that round-up or does it only round up if it is above .5?

naved001:
As per the python docs: https://docs.python.org/3/library/decimal.html#decimal.ROUND_HALF_UP

The effect is that, it rounds up if it equal to or above 5, otherwise it rounds it down.
e. g. `4.325` will be rounded to `4.33`. 

The default rounding method rounds to the nearest integer. 

default method:
```
>>> Decimal('4.325').quantize(Decimal(".01"))
Decimal('4.32')
>>> Decimal('4.315').quantize(Decimal(".01"))
Decimal('4.32')
```

ROUND_HALF_UP:
```
>>> Decimal('4.325').quantize(Decimal(".01"), rounding="ROUND_HALF_UP")
Decimal('4.33')
>>> Decimal('4.315').quantize(Decimal(".01"), rounding="ROUND_HALF_UP")
Decimal('4.32')
```

PR #36, opened 2024-05-01
Use 2 decimal points for Cost, rounding up

knikolla:
Switches the precision of Cost from 3 to 2. The value is rounded up as demontrated by the following block of code given the following csv file.

```csv
Cost
123.1234
123.5678
```

```python
>>> dataframe = pandas.read_csv("csv.csv", dtype={"Cost": pandas.ArrowDtype(pyarrow.decimal128(12, 2))})
>>> dataframe
     Cost
0  123.12
1  123.57
```

Related https://github.com/CCI-MOC/process_csv_report/issues/35

### DIFF
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -269,7 +269,7 @@ def merge_csv(files):
     dataframes = []
     for file in files:
         dataframe = pandas.read_csv(
-            file, dtype={COST_FIELD: pandas.ArrowDtype(pyarrow.decimal128(12, 3))}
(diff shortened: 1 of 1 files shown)


joachimweyl (approved):

naved001 (commented):
I think the rounding should probably happen in the source CSV reports. So, this PR should be okay as long as we agree on that.

naved001 (approved):

knikolla:
> Can we be explicit about the rounding method used?

@naved001 I surprisingly didn't find anything about how to change that or make it explicit.

Issue #43, opened 2024-05-07
Implement usage of PI alias file


QuanMPhm:
Because the current invoice processing script may create undesired output if a PI uses multiple aliases, we want to give the invoice script an alias file for it to correctly bill PIs with multiple aliases. This requires two separate tasks:

- [ ] Creation and storage of an alias file in long-term storage(?). Our list of currently known aliases are below. Hopefully no more aliases will appear the the future
- [ ] Modify `process_report.py` to use the alias file

QuanMPhm:
@knikolla I'll add an argument to `process_report.py` to accept a user-provided alias.csv file, with the option to also fetch the file from a hardcoded location in S3, similar to how the PI file will be fetched. 

Given an alias file, each time the invoice script applies the New-PI credit and makes the corresponding changes to the PI file, it will make sure changes to a PI's entry will be applied to the PI's aliases. 

To explain in more detail, as the invoice script [iterates through each billable PI](https://github.com/CCI-MOC/process_csv_report/blob/cd231b45086c8e5b97eae1e1f099b86759fc86b7/process_report/process_report.py#L349), determines how the New-PI credit should be applied to them, and modifies the PI's entry in the PI file, it make replicate that same modified entry across the PI's aliases.

This makes sure that in the (presumably) rare case where a PI appears in an invoice under more than 1 alias, the credits they use are still correctly tracked. 

@knikolla  Implementation of this alias file may be significantly more complex after #44 is merged. In which case #44 itself will need to be modified, or a new PR will be needed to update the alias file implementation?

knikolla:
@QuanMPhm The new argument makes sense.

With regards to the updating logic: 

A much simpler solution is to have a step before you start processing the new PI credit where you replace all aliases with the canonical name. 

So if your script encounters `username1` as a PI, and there is a line in alias for `username1 -> email1`, you replace username1 with email1 in all encountered rows of the file. 

This makes it so you don't have to replicate any changes to the PI file and you have one canonical row for each PI in the PI csv file.

QuanMPhm:
@knikolla Would we eventually remove alias entries from the PI file as they become out-of-date?

knikolla:
@QuanMPhm alias entries will not become out of date as users can't change their usernames.

Issue #41, opened 2024-05-03
Create a timestamped backup of PI.csv when the invoice processing script is ran


QuanMPhm:
Similar to how the processed invoices are timestamped and uploaded to a backup folder in S3, we want `PI.csv` to also be backed up

QuanMPhm:
@knikolla @joachimweyl  This functionally should be equivalent, but should I make the backup of the PI file **before** or **after** it has been modified?

knikolla:
> Before. Just incase something goes wrong with the previous run you know that a backup is made first thing.

++. Backing up data before you change it ensures you have a copy of the data which you can roll back to in case something goes wrong during processing.

PR #44, opened 2024-05-07
Added tracking of New-PI credit usage for second month

### LINKED ISSUE #37: Track usage of credit for a second month
## Motivation
Deal with the case where a user starts a project late in a month. We don't want them to loose out on the intro credit.

## Completion Criteria
automate tracking of new PIs so that we provide the 0002 credit for the first 2 months they show up on the invoice.

## Description
- [ ] Create a way to track the $1000 so that we know how much is left after the first month
   - new columns in the pi.csv might work to provide the amount and then we need to check the month associated and subtract the first month from the amount column and then, subtract the amount from the 2nd month and then never touch it again as they are past the 2 month mark
      - This will also allow for tracking of how much was used of the credit
      - 3 new columns in pi.csv amount, Initial Credit Amount - currently $1000, might change, after 1st month, after 2nd month
         - Initial Credit Amount is the credit offered at the time of the first month they were added to the list, it will not change in the future if the credit amount changes in the future.
            - Initial Credit Amount is the value for future runs if the script needs to be run again, that way if it changes in later months the new value is added to pi.csv and becomes the value that will always be used for that row
         - after 1st month = Initial Credit Amount - Cost for 1st month, down to min 0
         - after 2nd month = Initial Credit Amount - Cost for 1st month - Cost for 2nd month, down to min 0
      - make sure that rerunning the script does not create different values
- [ ] test that 2nd month the remaining value is used
   - test when more than the amount is used, what does the invoice show and what does the pi.csv show
   - test when less than the amount is used, what does the invoice show and what does the pi.csv show

### Example
1. PI1 joins in Jan uses $200
1. PI2 joins in Jan uses $600
1. PI3 joins in Feb and the credit amount drops to $500 and they use $200
1. PI1 uses $300 in Feb
1. PI2 uses $700 in Feb
1. PI3 uses $250 in Mar
1. pi.csv columns 
   1. PI1, 2024-01, 1000, 800, 500
   1. PI2, 2024-01, 1000, 400, 0
   1. PI3, 2024-02, 500, 300, 50
1. Invoice column PI, Month, Cost, credit, Balance
   1. PI1, 2024-01, 200, 200, 0
   1. PI2, 2024-01, 600, 600, 0
   1. PI1, 2024-02, 300, 300, 0
   1. PI2, 2024-02, 700, 400, 300
   1. PI3, 2024-02, 200, 200, 0
   1. PI3, 2024-03, 250, 250, 50

## Completion dates
Desired - 2024-05-29
Required - TBD

QuanMPhm:
Closes #37. I apologize for the large PR. I did not expect a feature summarized in 8 words would take this long...

Implementing second-month tracking requires the reformatting of the PI file, which I have created outside of this PR following the specifications in #37. That being said, the invoice processing script will now require a PI csv file containing the follow columns:
- PI
- First Invoice Month
- Initial Credits
- 1st Month Remainder
- 2nd Month Remainder

The most significant changes are the addition of some datetime utility functions, and mainly `apply_credits_new_pi()`. I've also re-written how the test case for applying the new-PI credit is done. Now the output invoice is directly compared to an answer Dataframe, and the modified PI file is directly compared to a list of strings. This should avoid the dizzying quantities of `self.assertEqual()` calls.

Given the combined invoice and the PI file, the function `apply_credits_new_pi()` will perform several steps as explained below:
1. Initialize the credit, credit code, and balance columns
2. Fetch the PI file, which is loaded into a dataframe `old_pi_df`
3. Get all 1-month PIs from the PI file and see if they have projects in the current invoice. If not, update the PI file as if they spent 0 credits
3. Obtain the invoice month, list of PIs for the current month, and attempt to get the initial credit amount for the current invoice using `old_pi_df`. If `old_pi_df` contains PIs for the current invoice month, and if those PIs have their initial credit amount set, then we use that value, otherwise, defaults to $1000
4. With each PI, determine the PI's 'age', which is the number of months since the PI's first invoice month until the current invoice month. This is 0 if the PI is first appeared on the current invoice, 1 if they appeared on the previous, and can be negative if we are processing past invoices.
5. With this 'age', the credit can be correctly applied to each PI, and the appropriate fields in `old_pi_df` are set.
6. Dump `old_pi_df` back into the pi file

I would like to note a few things when `apply_credits_new_pi()` applies the credits. @joachimweyl, you may find this most relevant?:
- If an invoice is modified and processed again, the updated PI file may change. Certain PIs may have their entries in the PI file overwritten. Information might not be lost since the script will also upload timestamped backups of invoices, and after #41, also the PI file. This may be relevant if we're processing past invoices, or if we decide to process different invoices for the same month before they are combined. Because the information needed to correctly apply the New-PI credit is centralized in one file, this means each month's invoice must be feed to the script at the same time. 
- When a PI first appears and a new entry is made the PI file, their `2nd Month Reminader` field will appear blank
- The script may give undesired output if PIs use aliases. I have made #43 to resolve this.

I have tried to be as comprehensive as possible with the test cases. I will add more per suggestion.

### DIFF
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -10,6 +10,15 @@ import boto3
 import pyarrow
 
 
+### PI file field names
+PI_PI_FIELD = "PI"
+PI_FIRST_MONTH = "First Invoice Month"
+PI_INITIAL_CREDITS = "Initial Credits"
+PI_1ST_USED = "1st Month Used"
+PI_2ND_USED = "2nd Month Used"
+###
+
+
 ### Invoice field names
 INVOICE_DATE_FIELD = "Invoice Month"
 PROJECT_FIELD = "Project - Allocation"
@@ -51,24 +60,31 @@ def load_institute_map() -> dict:
     return institute_map
 
 
-def load_old_pis(old_pi_file):
-    old_pi_dict = dict()
-
+def load_old_pis(old_pi_file) -> pandas.DataFrame:
     try:
-        with open(old_pi_file) as f:
-            for pi_info in f:
-                pi, first_month = pi_info.strip().split(",")
-                old_pi_dict[pi] = first_month
+        old_pi_df = pandas.read_csv(
+            old_pi_file,
+            dtype={
+                PI_INITIAL_CREDITS: pandas.ArrowDtype(pyarrow.decimal128(21, 2)),
+                PI_1ST_USED: pandas.ArrowDtype(pyarrow.decimal128(21, 2)),
+                PI_2ND_USED: pandas.ArrowDtype(pyarrow.decimal128(21, 2)),
+            },
+        )
     except FileNotFoundError:
         sys.exit("Applying credit 0002 failed. Old PI file does not exist")
 
-    return old_pi_dict
+    return old_pi_df
 
 
-def dump_old_pis(old_pi_file, old_pi_dict: dict):
-    with open(old_pi_file, "w") as f:
-        for pi, first_month in old_pi_dict.items():
-            f.write(f"{pi},{first_month}\n")
+def dump_old_pis(old_pi_file, old_pi_df: pandas.DataFrame):
+    old_pi_df = old_pi_df.astype(
+        {
+            PI_INITIAL_CREDITS: pandas.ArrowDtype(pyarrow.decimal128(21, 2)),
+            PI_1ST_USED: pandas.ArrowDtype(pyarrow.decimal128(21, 2)),
+            PI_2ND_USED: pandas.ArrowDtype(pyarrow.decimal128(21, 2)),
+        },
+    )
+    old_pi_df.to_csv(old_pi_file, index=False)
 
 
 def load_alias(alias_file):
@@ -86,22 +102,29 @@ def load_alias(alias_file):
     return alias_dict
 
 
-def is_old_pi(old_pi_dict, pi, invoice_month):
-    first_invoice_month = old_pi_dict.get(pi, invoice_month)
-    if compare_invoice_month(first_invoice_month, invoice_month):
+def get_pi_age(old_pi_df: pandas.DataFrame, pi, invoice_month):
+    """Returns time difference between current invoice month and PI's first invoice month
+    I.e 0 for new PIs
+
+    Will raise an error if the PI'a age is negative, which suggests a faulty invoice, or a program bug"""
+    first_invoice_month = old_pi_df.loc[old_pi_df[PI_PI_FIELD] == pi, PI_FIRST_MONTH]
+    if first_invoice_month.empty:
+        return 0
+
+    month_diff = get_month_diff(invoice_month, first_invoice_month.iat[0])
+    if month_diff < 0:
         sys.exit(
             f"PI {pi} from {first_invoice_month} found in {invoice_month} invoice!"
         )
-    if compare_invoice_month(invoice_month, first_invoice_month):
-        return True
-    return False
+    else:
+        return month_diff
 
 
-def compare_invoice_month(month_1, month_2):
-    """Returns True if 1st date is later than 2nd date"""
+def get_month_diff(month_1, month_2):
+    """Returns a positive integer if month_1 is ahead in time of month_2"""
     dt1 = datetime.datetime.strptime(month_1, "%Y-%m")
     dt2 = datetime.datetime.strptime(month_2, "%Y-%m")
-    return dt1 > dt2
+    return (dt1.year - dt2.year) * 12 + (dt1.month - dt2.month)
 
 
 def get_invoice_bucket():
@@ -384,28 +407,57 @@ def fetch_s3_alias_file():
 
 def apply_credits_new_pi(dataframe, old_pi_file):
     new_pi_credit_code = "0002"
-    new_pi_credit_amount = 1000
+    INITIAL_CREDIT_AMOUNT = 1000
     EXCLUDE_SU_TYPES = ["OpenShift GPUA100SXM4", "OpenStack GPUA100SXM4"]
 
     dataframe[CREDIT_FIELD] = None
     dataframe[CREDIT_CODE_FIELD] = None
     dataframe[BALANCE_FIELD] = Decimal(0)
 
-    old_pi_dict = load_old_pis(old_pi_file)
+    old_pi_df = load_old_pis(old_pi_file)
 
-    current_pi_list = dataframe[PI_FIELD].unique()
+    current_pi_set = set(dataframe[PI_FIELD])
     invoice_month = dataframe[INVOICE_DATE_FIELD].iat[0]
+    invoice_pis = old_pi_df[old_pi_df[PI_FIRST_MONTH] == invoice_month]
+    if invoice_pis[PI_INITIAL_CREDITS].empty or pandas.isna(
+        new_pi_credit_amount := invoice_pis[PI_INITIAL_CREDITS].iat[0]
+    ):
+        new_pi_credit_amount = INITIAL_CREDIT_AMOUNT
 
-    for pi in current_pi_list:
+    print(f"New PI Credit set at {new_pi_credit_amount} for {invoice_month}")
+
+    for pi in current_pi_set:
         pi_projects = dataframe[dataframe[PI_FIELD] == pi]
+        pi_age = get_pi_age(old_pi_df, pi, invoice_month)
+        pi_old_pi_entry = old_pi_df.loc[old_pi_df[PI_PI_FIELD] == pi].squeeze()
 
-        if is_old_pi(old_pi_dict, pi, invoice_month):
+        if pi_age > 1:
             for i, row in pi_projects.iterrows():
                 dataframe.at[i, BALANCE_FIELD] = row[COST_FIELD]
         else:
-            old_pi_dict[pi] = invoice_month
-            print(f"Found new PI {pi}")
-            remaining_credit = new_pi_credit_amount
+            if pi_age == 0:
+                if len(pi_old_pi_entry) == 0:
+                    pi_entry = [pi, invoice_month, new_pi_credit_amount, 0, 0]
+                    old_pi_df = pandas.concat(
+                        [
+                            pandas.DataFrame([pi_entry], columns=old_pi_df.columns),
+                            old_pi_df,
+                        ],
+                        ignore_index=True,
+                    )
+                    pi_old_pi_entry = old_pi_df.loc[
+                        old_pi_df[PI_PI_FIELD] == pi
+                    ].squeeze()
+
+                remaining_credit = new_pi_credit_amount
+                credit_used_field = PI_1ST_USED
+            elif pi_age == 1:
+                remaining_credit = (
+                    pi_old_pi_entry[PI_INITIAL_CREDITS] - pi_old_pi_entry[PI_1ST_USED]
+                )
+                credit_used_field = PI_2ND_USED
+
+            initial_credit = remaining_credit
             for i, row in pi_projects.iterrows():
                 if remaining_credit == 0 or row[SU_TYPE_FIELD] in EXCLUDE_SU_TYPES:
                     dataframe.at[i, BALANCE_FIELD] = row[COST_FIELD]
@@ -418,7 +470,18 @@ def apply_credits_new_pi(dataframe, old_pi_file):
                     dataframe.at[i, BALANCE_FIELD] = row[COST_FIELD] - applied_credit
                     remaining_credit -= applied_credit
 
-    dump_old_pis(old_pi_file, old_pi_dict)
+            credits_used = initial_credit - remaining_credit
+            if (pi_old_pi_entry[credit_used_field] != 0) and (
+                credits_used != pi_old_pi_entry[credit_used_field]
+            ):
+                print(
+                    f"Warning: PI file overwritten. PI {pi} previously used ${pi_old_pi_entry[credit_used_field]} of New PI credits, now uses ${credits_used}"
+                )
+            old_pi_df.loc[
+                old_pi_df[PI_PI_FIELD] == pi, credit_used_field
+            ] = credits_used
+
+    dump_old_pis(old_pi_file, old_pi_df)
 
     return dataframe
 
--- a/process_report/tests/unit_tests.py
+++ b/process_report/tests/unit_tests.py
@@ -1,6 +1,7 @@
 from unittest import TestCase, mock
 import tempfile
 import pandas
+import pyarrow
 import os
 import math
 from textwrap import dedent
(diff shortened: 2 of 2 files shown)


process_report/process_report.py
```diff
@@ -46,43 +55,53 @@ def load_institute_map() -> dict:
...
+    old_pi_df.to_csv(old_pi_file, index=False)
 
 
-def is_old_pi(old_pi_dict, pi, invoice_month):
-    first_invoice_month = old_pi_dict.get(pi, invoice_month)
-    if compare_invoice_month(first_invoice_month, invoice_month):
+def get_pi_age(old_pi_df: pandas.DataFrame, pi, invoice_month):
+    """Returns time difference between current invoice month and PI's first invoice month
+    I.e 0 for new PIs
+    Can be negative if processing past invoices"""
```
naved001, line 75: but it raises an error when it's negative

process_report/process_report.py
```diff
@@ -46,43 +55,53 @@ def load_institute_map() -> dict:
...
+
+
+def add_invoice_month(invoice_month, month_to_add):
+    """Given an invoice month string, i.e "2023-02", and an integer,
+    add as many months to the invoice month, and return the sum in same
+    format"""
+    dt = datetime.datetime.strptime(invoice_month, "%Y-%m")
+    new_month = dt.month - 1 + month_to_add
+    new_year = dt.year + new_month // 12
+    new_dt = datetime.datetime(year=new_year, month=new_month % 12 + 1, day=1)
```
naved001, line 103: it's not intuitive as to why we need to subtract 1 in line 101 and then add 1 back again.

process_report/process_report.py
```diff
@@ -335,40 +354,76 @@ def validate_pi_names(dataframe):
...
+    print(f"New PI Credit set at {new_pi_credit_amount} for {invoice_month}")
+
+    one_month_pi_list = old_pi_df[
+        old_pi_df[PI_FIRST_MONTH] == add_invoice_month(invoice_month, -1)
+    ][PI_PI_FIELD]
+    for pi in one_month_pi_list:
+        if pi not in current_pi_set:
+            old_pi_df.loc[
+                old_pi_df[PI_PI_FIELD] == pi, PI_2ND_REMAINDER
+            ] = old_pi_df.loc[old_pi_df[PI_PI_FIELD] == pi, PI_1ST_REMAINDER].iat[0]
```
naved001, line 381: is this possible to break it down into smaller steps?

naved001 (commented):
I need to take another look at this, it's kind of hard to follow the logic

process_report/process_report.py
QuanMPhm: I should have wrote the docstring a bit clearer. When writing this, I believed it should be intended behavior for the program to raise an error if the program was processing a past invoice, but encountered a PI that was suppose to appear in the future. 

I.e if the PI file had an entry for PI A for 2024-01, but PI A was found when processing past invoices, an error should be raised since that might suggest the invoice was wrong, or the program was faulty in the past.

I will remove the ` Can be negative if processing past invoices` comment from the docstring

process_report/process_report.py
QuanMPhm: I didn't come up with the math for this myself, this was taken from this [StackOverflow answer](https://stackoverflow.com/a/4131114). I have added a few more testcases for this function to be sure.

process_report/process_report.py
QuanMPhm: Turns out all this could be a two-liner. Also, because of your suggestion,  when I tested the testcases I realized there was a test case that made the `Apply New PI Credit` test always pass.

process_report/process_report.py
```diff
@@ -46,43 +55,54 @@ def load_institute_map() -> dict:
...
-def is_old_pi(old_pi_dict, pi, invoice_month):
-    first_invoice_month = old_pi_dict.get(pi, invoice_month)
-    if compare_invoice_month(first_invoice_month, invoice_month):
+def get_pi_age(old_pi_df: pandas.DataFrame, pi, invoice_month):
+    """Returns time difference between current invoice month and PI's first invoice month
+    I.e 0 for new PIs
+
+    Will raise an error if the PI'a age is negative, which suggests a faulty invoice, or a program bug"""
+    first_invoice_month = old_pi_df.loc[old_pi_df[PI_PI_FIELD] == pi, PI_FIRST_MONTH]
+    if first_invoice_month.empty:
```
knikolla, line 78: This should also throw up an error, I can't see a reason why the first invoice month for a PI would be empty in the CSV and why that would be okay.

process_report/process_report.py
knikolla: I don't understand why this is necessary? This calculation is only used for your bookkeeping and can be inferred from the age of the PI. If pi age is 0, don't read remainder, write remainder in first remainder column. if pi age is 1, read first remainder, write second remainder.

knikolla (changes requested):

process_report/process_report.py
QuanMPhm: This would be for the case where the PI doesn't have an entry in the PI file yet, so they are a completely new PI.

knikolla: Ah, I see. This is okay then.

process_report/process_report.py
QuanMPhm: I guess I didn't want to iterate through the old_pi dataframe to check for PIs with an age of 1. I thought writing a function to calculate and return the invoice month string would make the code more organized. Also, you said:
> If pi age is 0, don't read remainder, write remainder in first remainder column. if pi age is 1, read first remainder, write second remainder.

Just so I'm sure, you want it so that, if the pi age is 0 (new PI), and they have an entry in the PI file already (maybe because we're re-processing same invoice, different invoice for same month, etc), you want their entry to be overwritten and possibly change?

knikolla: > I guess I didn't want to iterate through the old_pi dataframe to check for PIs with an age of 1. I thought writing a function to calculate and return the invoice month string would make the code more organized. Also, you said:

Why do you need to iterate through the old_pi dataframe to check for PIs with an age of 1?

For each PI that you are processing you can find the entry in the PI file and have all the information you need. 

> 
> > If pi age is 0, don't read remainder, write remainder in first remainder column. if pi age is 1, read first remainder, write second remainder.
> 
> Just so I'm sure, you want it so that, if the pi age is 0 (new PI), and they have an entry in the PI file already (maybe because we're re-processing same invoice, different invoice for same month, etc), you want their entry to be overwritten and possibly change?

Yes, do overwrite it and print a warning if the value is different.

QuanMPhm: > Why do you need to iterate through the old_pi dataframe to check for PIs with an age of 1?

I needed to iterate through the PI file instead of the current invoice to find these 1-month-old PIs because they might not appear (have billable projects) in the current month's invoice. If the 1-month-old PI doesn't appear in the current invoice, it's logically means they have used $0 of credit, and this is the reason why this [code block](https://github.com/CCI-MOC/process_csv_report/pull/44/files#diff-162d25340846c53be6b9c357f242c269b4564a175d9f873c6a37a63239dd5419R375-R378) (which is why `add_invoice_month()` is made) was added.

I will add warnings in the code for the warning, and now that I think about it, do you want me to using the `logging` module, instead of `print()`, or should that be a seperate PR?

knikolla: Ok, I understand the reasoning. For PIs on their second month (age == 1) and have no usage, you want to populate the value of the second remainder correctly.

I think it's a fair trade off to just leave those values Null for the sake of code simplicity. The code is already hard to follow and the value for second remainder is not used by anything.

- For each PI in current invoice
  - If age of PI == 0
    - Do calculation as before. At the end create new row in PIs.csv with current invoice month, credit and remainder of first month. Value for remainder of second month is empty.
  - If age of PI == 1
    - Read the remainder of second month from PIs.csv. Write remainder of second month. 
    - If no usage in second month, second remainder will be a null value.

knikolla: > do you want me to using the `logging` module, instead of `print()`, or should that be a seperate PR?

that can be a separate PR.

knikolla: > Ok, I understand the reasoning. For PIs on their second month (age == 1) and have no usage, you want to populate the value of the second remainder correctly.
> 
> I think it's a fair trade off to just leave those values Null for the sake of code simplicity. The code is already hard to follow and the value for second remainder is not used by anything.
> 
> * For each PI in current invoice
>   
>   * If age of PI == 0
>     
>     * Do calculation as before. At the end create new row in PIs.csv with current invoice month, credit and remainder of first month. Value for remainder of second month is empty.
>   * If age of PI == 1
>     
>     * Read the remainder of second month from PIs.csv. Write remainder of second month.
>     * If no usage in second month, second remainder will be a null value.

Alternatively, and what I proposed before and I favor, is instead of counting the `remainder` of the credit to instead use that column to track `credit used` and preinitialize both first and second month to 0 and update them as you go.

joachimweyl: I believe that is why we have a column to track the original value so we can calculate the remainder by subtracting the credit used from the original credit to get the credit remaining.

QuanMPhm: @knikolla To be sure...
> Alternatively, and what I proposed before and I favor, 
Does this mean you favor your second suggestion, or first one, or both?

> If no usage in second month, second remainder will be a null value.
With the current code, whenever a new PI entry is added, their `1st Month Remainder` and `2nd Month Remainder` fields are initialized to `None` (which is empty when written out to a csv file). I take it that you don't believe I should include [this code block](https://github.com/CCI-MOC/process_csv_report/pull/44/files#diff-162d25340846c53be6b9c357f242c269b4564a175d9f873c6a37a63239dd5419R375-R378) because there's already so much going on, so I'll just remove it then. With this block removed, 1-month PIs who have no usage in the second month will have their `2nd Month Remainder` set as empty.

> preinitialize both first and second month to 0 and update them as you go.
I think if the `1st Month Remainder` and `2nd Month Remainder` fields are initialized to 0, there will be ambiguity on whether a PI spent all their credits, or if they didn't appear on their second month of invoicing?

knikolla: I was saying that's why in my initial suggestion when we were discussion the feature implementation I favor the column to mean Credit Used (1st and second month) instead of Credit Remaining (first and second month), in which case Credit Used zero would be unambiguous.

I'm okay with keeping the columns to mean Credit Remainder 1st month and Credit remainder 2nd month and for an empty value to mean credit not used.

QuanMPhm: Could I ask why having the columns to track `Credits Used` instead of `Credits Remaining` would be better? I will follow your preference, since I don't see much difference in terms of structure or burden of maintenance for either approach.

knikolla: Mostly a gut feeling, but if I have to try to think about that preference a bit more.

- The values of credit used for the first month and the second month are independent.
- They hold meaning by themselves, as the value that we forgave from users balances for their 1st and 2nd month.
- In most cases they will match the balance for their first and (sometimes) second month unless big expenditures.
- I think of the credit used as the "active" number that is decreasing the value of the starting credit (static). So instead of writing down 1000, 400, 200 (the sequence of changing state). In my head I can better visualize it when written as 1000, -600, -200 (balance and the 2 expenditures).

knikolla:
@QuanMPhm hey Quan, are you still working on this? I see that some comments haven't been resolved per discussion.

QuanMPhm:
@knikolla Oh sorry, I should have pinged you. I made changes according to your comments [here](https://github.com/CCI-MOC/process_csv_report/pull/44#discussion_r1604120234) and [here](https://github.com/CCI-MOC/process_csv_report/pull/44#discussion_r1603705500). Sorry again

process_report/process_report.py
```diff
@@ -347,21 +367,50 @@ def apply_credits_new_pi(dataframe, old_pi_file):
...
     dataframe[BALANCE_FIELD] = Decimal(0)
 
-    old_pi_dict = load_old_pis(old_pi_file)
+    old_pi_df = load_old_pis(old_pi_file)
 
-    current_pi_list = dataframe[PI_FIELD].unique()
+    current_pi_set = set(dataframe[PI_FIELD])
     invoice_month = dataframe[INVOICE_DATE_FIELD].iat[0]
+    invoice_pis = old_pi_df[old_pi_df[PI_FIRST_MONTH] == invoice_month]
+    if invoice_pis["Initial Credits"].empty or pandas.isna(
```
knikolla, line 375: you have defined the string as a global variable, use that instead.

process_report/process_report.py
```diff
@@ -347,21 +367,50 @@ def apply_credits_new_pi(dataframe, old_pi_file):
...
+        new_pi_credit_amount := invoice_pis["Initial Credits"].iat[0]
+    ):
+        new_pi_credit_amount = 1000
```
knikolla, line 378: You've already defined `new_pi_credit_amount` as 1000 at the top of the function. Either use a different variable name to prevent overwriting it, or do not set this variable to a different value if there is not a number in that column.

process_report/process_report.py
```diff
@@ -48,43 +57,54 @@ def load_institute_map() -> dict:
...
-def load_old_pis(old_pi_file):
-    old_pi_dict = dict()
-
+def load_old_pis(old_pi_file) -> pandas.DataFrame:
     try:
-        with open(old_pi_file) as f:
-            for pi_info in f:
-                pi, first_month = pi_info.strip().split(",")
-                old_pi_dict[pi] = first_month
+        old_pi_df = pandas.read_csv(old_pi_file)
```
knikolla, line 62: You're not loading this as Decimal values. See `merge_csv` function on how.

process_report/process_report.py
knikolla: Please remove this function (and its associated test) since we're not using it anymore.

knikolla (changes requested):
Not fully finished reviewing but posting this before I break for lunch.

process_report/tests/unit_tests.py
```diff
@@ -354,39 +506,13 @@ def test_apply_credit_0002(self):
         dataframe = process_report.apply_credits_new_pi(
             self.dataframe, self.old_pi_file
         )
+        dataframe = dataframe.astype({"Credit": "float64", "Balance": "int64"})
```
knikolla, line 509: Only using integers and floats makes this test not representative. Everything should be `Decimal`. But this can be changed in a follow-up.

knikolla (changes requested):
Went through the test case and the values are correct for the behavior we want. There's a lot of opportunities to split the apply credits function into smaller chunks for easier readability, testability and maintainability but this can be done in a follow-up patch since we need this for this months billing cycle.

QuanMPhm:
@knikolla @naved001  I've rebased the PR, and should have resolved everything except for refactoring the Test Cases

process_report/process_report.py
```diff
@@ -51,24 +60,24 @@ def load_institute_map() -> dict:
...
-
+def load_old_pis(old_pi_file) -> pandas.DataFrame:
     try:
-        with open(old_pi_file) as f:
-            for pi_info in f:
-                pi, first_month = pi_info.strip().split(",")
-                old_pi_dict[pi] = first_month
+        old_pi_df = pandas.read_csv(
+            old_pi_file,
+            converters={
```
knikolla, line 67: why are you using converters instead of specifying a `dtype` as in the `merge_csv` function?

QuanMPhm: I used converters because specifying `decimal.Decimal` as a dtype would raise the error:
`TypeError: dtype '<class 'decimal.Decimal'>' not understood`
From your comment, I digged at bit more, and it seems `decimal.Decimal` simply isn't a type [actively supported](https://github.com/pandas-dev/pandas/issues/18991#issuecomment-354478790) by the Pandas module, so we must specify it as a converter, not dtype.

In any case, I made sure that the types of the columns were `decimal.Decimal`

knikolla: That's what `pyarrow` provides. A Decimal `dtype`. 

https://github.com/CCI-MOC/process_csv_report/blob/43d43410850a3e6fc752ccec0ad0119a8113dbd5/process_report/process_report.py#L307

knikolla (approved):

knikolla:
@naved001 do you have any further comments?

naved001 (approved):

PR #47, opened 2024-05-07
Upload backup of PI file after processing

### LINKED ISSUE #41: Create a timestamped backup of PI.csv when the invoice processing script is ran

QuanMPhm:
Closes #41 `process_report.py` will now upload a backup of the PI file before it is modified during invoice processing. This is done in `fetch_s3_old_pi_file()` when the PI file is first fetched, and is noted with a comment for clarification.

### DIFF
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -230,6 +230,9 @@ def main():
 
     billable_projects = remove_non_billables(merged_dataframe, pi, projects)
     billable_projects = validate_pi_names(billable_projects)
+
+    if args.upload_to_s3:
+        backup_to_s3_old_pi_file(old_pi_file)
     credited_projects = apply_credits_new_pi(billable_projects, old_pi_file)
 
     export_billables(credited_projects, args.output_file)
@@ -385,6 +388,11 @@ def fetch_s3_old_pi_file():
 def upload_to_s3_old_pi_file(old_pi_file):
     invoice_bucket = get_invoice_bucket()
     invoice_bucket.upload_file(old_pi_file, PI_S3_FILEPATH)
+
+
+def backup_to_s3_old_pi_file(old_pi_file):
+    invoice_bucket = get_invoice_bucket()
+    invoice_bucket.upload_file(old_pi_file, f"PIs/Archive/PI {get_iso8601_time()}.csv")
 
 
 def add_institution(dataframe: pandas.DataFrame):


process_report/process_report.py
```diff
@@ -377,6 +377,9 @@ def fetch_s3_old_pi_file():
     local_name = "PI.csv"
     invoice_bucket = get_invoice_bucket()
     invoice_bucket.download_file("PIs/PI.csv", local_name)
+    invoice_bucket.upload_file(
+        local_name, f"PIs/Archive/PI {get_iso8601_time()}.csv"
```
larsks, line 381: Not a big deal, but it's generally best to avoid spaces in filenames (there are a variety of standard tools that need to be explicitly configured to handle whitespace in filenames).

process_report/process_report.py
```diff
@@ -377,6 +377,9 @@ def fetch_s3_old_pi_file():
...
+    )  # Create backup when fetching
```
naved001, line 382: This will try to always create this backup copy every time this program is run. We call `fetch_s3_old_pi_file` very early in the program, basically right after we parse the arguments this call is made; so even before the program does anything useful it is called.

This also ignores the `--upload-to-s3` flag because.

I think the best time to upload the backup/old pi file would be right before you create the updated version.

QuanMPhm: I have made it such that the PI file will be backed up right before the New PI credit is applied.

knikolla (approved):

process_report/process_report.py
```diff
@@ -230,6 +230,8 @@ def main():
 
     billable_projects = remove_non_billables(merged_dataframe, pi, projects)
     billable_projects = validate_pi_names(billable_projects)
+
+    backup_to_s3_old_pi_file(old_pi_file)
```
naved001, line 234: you should check if `args.upload_to_s3` is set before trying to upload to s3 because otherwise if I make test runs I'd be uploading unnecessary stuff to S3 which I'd have to then clean up.

process_report/process_report.py
```diff
@@ -387,6 +389,11 @@ def upload_to_s3_old_pi_file(old_pi_file):
     invoice_bucket.upload_file(old_pi_file, PI_S3_FILEPATH)
 
 
+def backup_to_s3_old_pi_file(old_pi_file):
+    invoice_bucket = get_invoice_bucket()
+    invoice_bucket.upload_file(old_pi_file, f"PIs/Archive/PI {get_iso8601_time()}.csv")
```
naved001, line 394: `backup_to_s3_old_pi_file` and `upload_to_s3_old_pi_file` can be one method and you can just pass the s3 filepath as an argument, but you can do that as part of refactoring.

QuanMPhm: I'll do that as part of refactoring later.

naved001 (approved):

PR #50, opened 2024-05-09
Validate combined dataframe with alias file

### LINKED ISSUE #43: Implement usage of PI alias file

QuanMPhm:
Closes #43. Added an argument `alias-file` which gives the user the choice to provide a PI alias file, else defaults to fetching the file from a hardcoded location in S3 storage ("PIs/alias.csv")

The PI file must be a csv, with the first value of each row being the canonical name for each PI every other value in each row will be the PI's known aliases.

An example:
```
PI1,PI1_1,PI1_2
PI2,PI2_1
```

Given the alias file, `validate_pi_aliases` will iterate through all projects in the combined invoice, and replace all encountered aliases with its PI's canonical name

### DIFF
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -31,6 +31,9 @@ BALANCE_FIELD = "Balance"
 PI_S3_FILEPATH = "PIs/PI.csv"
 
 
+ALIAS_S3_FILEPATH = "PIs/alias.csv"
+
+
 def get_institution_from_pi(institute_map, pi_uname):
     institution_key = pi_uname.split("@")[-1]
     institution_name = institute_map.get(institution_key, "")
@@ -57,8 +60,7 @@ def load_old_pis(old_pi_file):
                 pi, first_month = pi_info.strip().split(",")
                 old_pi_dict[pi] = first_month
     except FileNotFoundError:
-        print("Applying credit 0002 failed. Old PI file does not exist")
-        sys.exit(1)
+        sys.exit("Applying credit 0002 failed. Old PI file does not exist")
 
     return old_pi_dict
 
@@ -67,6 +69,21 @@ def dump_old_pis(old_pi_file, old_pi_dict: dict):
     with open(old_pi_file, "w") as f:
         for pi, first_month in old_pi_dict.items():
             f.write(f"{pi},{first_month}\n")
+
+
+def load_alias(alias_file):
+    alias_dict = dict()
+
+    try:
+        with open(alias_file) as f:
+            for line in f:
+                pi_alias_info = line.strip().split(",")
+                alias_dict[pi_alias_info[0]] = pi_alias_info[1:]
+    except FileNotFoundError:
+        print("Validating PI aliases failed. Alias file does not exist")
+        sys.exit(1)
+
+    return alias_dict
 
 
 def is_old_pi(old_pi_dict, pi, invoice_month):
@@ -189,6 +206,11 @@ def main():
         help="Name of csv file listing previously billed PIs. If not provided, defaults to fetching from S3",
     )
     parser.add_argument(
+        "--alias-file",
+        required=False,
+        help="Name of alias file listing PIs with aliases (and their aliases). If not provided, defaults to fetching from S3",
+    )
+    parser.add_argument(
         "--BU-subsidy-amount",
         required=True,
         type=int,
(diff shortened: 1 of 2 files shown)


process_report/process_report.py
```diff
@@ -333,6 +364,23 @@ def validate_pi_names(dataframe):
...
+            dataframe.loc[dataframe[PI_FIELD] == pi_alias, PI_FIELD] = pi
+
+    return dataframe
+
+
+def fetch_s3_alias_file():
+    local_name = "alias.csv"
+    invoice_bucket = get_invoice_bucket()
+    invoice_bucket.download_file(ALIAS_S3_FILEPATH, local_name)
+    return local_name
```
larsks, line 381: It's a little weird that we're returning a static filename to the caller. Should this just be a global constant instead?

process_report/process_report.py
```diff
@@ -67,6 +70,21 @@ def dump_old_pis(old_pi_file, old_pi_dict: dict):
...
+    try:
+        with open(alias_file) as f:
+            for line in f:
+                pi_alias_info = line.strip().split(",")
+                alias_dict[pi_alias_info[0]] = pi_alias_info[1:]
+    except FileNotFoundError:
+        print("Validating PI aliases failed. Alias file does not exist")
+        sys.exit(1)
+
+    return alias_dict
```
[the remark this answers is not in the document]
QuanMPhm: Note taken!

process_report/process_report.py
QuanMPhm: I only use it once in this function, so I didn't thought it would be necessary to make it a global. Should I?

larsks: I guess my question is why we're bothing to return a value here at all, since it's always the same thing.

QuanMPhm: I...um...don't have a strong reason why it's like that? I guess I just wanted all the related code in one place, but if you think it's better to make this a constant global, I'll do so.

larsks: I don't have a strong opinion here. It's fine if you leave it the way it is for now; we can always change it later if we develop a compelling reason to do so.

process_report/process_report.py
```diff
@@ -335,6 +366,24 @@ def validate_pi_names(dataframe):
     return dataframe
 
 
+def validate_pi_aliases(dataframe: pandas.DataFrame, alias_dict: dict):
+    for pi, pi_aliases in alias_dict.items():
+        for pi_alias in pi_aliases:
+            if not dataframe.loc[dataframe[PI_FIELD] == pi_alias].empty:
+                print(f"Warning: Found alias {pi_alias} for {pi}")
+
+        dataframe.loc[dataframe[PI_FIELD].isin(pi_aliases), PI_FIELD] = pi
```
larsks, line 375: The entire point in using the vector operations provided by pandas is to avoid performing iterations in Python. Just get rid of the inner for loop.

```suggestion
    for pi, pi_aliases in alias_dict.items():
        dataframe.loc[dataframe[PI_FIELD].isin(pi_aliases), PI_FIELD] = pi
```

QuanMPhm: I kept the inner loop because I wanted to log whether a PI's alias was encountered. I don't think there is a way to do this without using iterating through the PI's aliases?

knikolla: It's okay not to log when alias encountered.

knikolla (approved):

process_report/process_report.py
```diff
@@ -69,6 +72,21 @@ def dump_old_pis(old_pi_file, old_pi_dict: dict):
...
+    alias_dict = dict()
+
+    try:
+        with open(alias_file) as f:
+            for line in f:
+                pi_alias_info = line.strip().split(",")
+                alias_dict[pi_alias_info[0]] = pi_alias_info[1:]
+    except FileNotFoundError:
+        print("Validating PI aliases failed. Alias file does not exist")
+        sys.exit(1)
```
naved001, line 85: I think this was already pointed out, but this should be 
```python
sys.exit("Validating PI aliases failed. Alias file does not exist")
```

naved001 (approved):

PR #54, opened 2024-05-23
Initial refactoring of process_report

### LINKED ISSUE #63: Refactor `process_csv_report.py` by restructuring code, and refactor Lenovo and nonbillable invoices
Split up from #52. We want to refactor the current code such that every invoice will implement a common base class, `Invoice`. This would allow for a more modular codebase that is easier to review and test.

The first step of refactoring will be to create the base class, and refactor the Lenovo and nonbillable invoices.

QuanMPhm:
Closes #63 . This refactor commit is the first of a few, to lay out initial structure

A new submodule, `invoices`, is added, containing a base class `Invoice` which is inherited by all other invoices. Currently, only the lenovo and nonbillable invoice has classes which inherits from `Invoice` Also created and partially populated an `util.py` file, containing functions placed above `main()` in `process_report.py` After the refactoring process is fully complete,
these utility functions will be completely removed from `process_report.py`

I can already tell I am breaking several Python conventions, so I am ready.

### DIFF
--- a/process_report/invoices/nonbillable_invoice.py
+++ b/process_report/invoices/nonbillable_invoice.py
@@ -1,0 +1,15 @@
+from dataclasses import dataclass
+
+import process_report.invoices.invoice as invoice
+
+
+@dataclass
+class NonbillableInvoice(invoice.Invoice):
+    nonbillable_pis: list[str]
+    nonbillable_projects: list[str]
+
+    def _prepare_export(self):
+        self.data = self.data[
+            self.data[invoice.PI_FIELD].isin(self.nonbillable_pis)
+            | self.data[invoice.PROJECT_FIELD].isin(self.nonbillable_projects)
+        ]
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -9,6 +9,8 @@ import pandas
 import boto3
 import pyarrow
 
+from process_report.invoices import lenovo_invoice, nonbillable_invoice
+
 
 ### Invoice field names
 INVOICE_DATE_FIELD = "Invoice Month"
@@ -150,7 +152,7 @@ def main():
     parser.add_argument(
         "--nonbillable-file",
         required=False,
-        default="nonbillable.csv",
+        default="nonbillable",
         help="Name of nonbillable file",
     )
     parser.add_argument(
@@ -180,7 +182,7 @@ def main():
     parser.add_argument(
         "--Lenovo-file",
         required=False,
-        default="Lenovo.csv",
+        default="Lenovo",
         help="Name of output csv for Lenovo SU Types invoice",
     )
     parser.add_argument(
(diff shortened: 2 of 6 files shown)


process_report/util.py
```diff
@@ -0,0 +1,30 @@
...
+
+def load_institute_map() -> dict:
+    with open("process_report/institute_map.json", "r") as f:
+        institute_map = json.load(f)
+
+    return institute_map
+
+
+def get_iso8601_time():
+    return datetime.datetime.now().strftime("%Y%m%dT%H%M%SZ")
```
larsks, line 23: Why not just use the `datetime.datetime.isoformat` method?

```
>>> datetime.datetime.now().isoformat(timespec='seconds')
'2024-05-24T13:16:40'
```

You get delimiters in the date, but otherwise the format is identical.

```suggestion
def get_iso8601_time():
    return datetime.datetime.now().isoformat(timespec='seconds')
```

process_report/process_report.py
```diff
@@ -221,12 +223,21 @@ def main():
...
     print("The following timed-projects will not be billed for this period: ")
     print(timed_projects_list)
-
     projects = list(set(projects + timed_projects_list))
 
     merged_dataframe = add_institution(merged_dataframe)
-    export_lenovo(merged_dataframe, args.Lenovo_file)
-    remove_billables(merged_dataframe, pi, projects, args.nonbillable_file)
+    lenovo_inv = lenovo_invoice.LenovoInvoice(
+        args.Lenovo_file, invoice_month, merged_dataframe.copy()
```
knikolla, line 230: For both these calls to constructors for the invoices I would advise you use named parameters. This because it's hard to trace the order of arguments in the subclasses without the lack of a `__init__()` function to look at, since this order is dependent on the order of definition of attributes of the dataclass. 

```suggestion
    lenovo_inv = lenovo_invoice.LenovoInvoice(
         name=args.Lenovo_file, invoice_month=invoice_month, data=merged_dataframe.copy()
         )
```

process_report/util.py
```diff
@@ -0,0 +1,30 @@
+import datetime
+import json
+
+
+def get_institution_from_pi(institute_map, pi_uname):
+    institution_key = pi_uname.split("@")[-1]
+    institution_name = institute_map.get(institution_key, "")
+
+    if institution_name == "":
+        print(f"Warning: PI name {pi_uname} does not match any institution!")
```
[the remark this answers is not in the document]
QuanMPhm: @knikolla Should this become its own PR, or do I make use of `logging` in this intial refactor PR?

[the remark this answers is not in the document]
knikolla: @QuanMPhm For the specific invoices that you are refactoring in this PR, make use of logging.

process_report/invoices/nonbillable_invoice.py
```diff
@@ -0,0 +1,15 @@
...
+@dataclass
+class NonbillableInvoice(invoice.Invoice):
+    nonbillable_pis: list[str]
+    nonbillable_projects: list[str]
+
+    def _prepare_export(self):
+        self.data = self.data[
+            self.data[invoice.PI_FIELD].isin(self.nonbillable_pis)
+            | self.data[invoice.PROJECT_FIELD].isin(self.nonbillable_projects)
+        ]
```
naved001, line 15: how come we don't implement a ` _process` for this class?

QuanMPhm: The only thing needed to make the nonbillable invoice from the merged csv was to filter out nonbillable PIs and projects. I thought this would fit into the `_prepare_export` function instead of `_process`, since there wasn't any real computations being performed, only basic filtering. 

I guess this boiled down to a small design choice for me.

naved001: @QuanMPhm you are absolutely right, we only need to merge so there's nothing to process.

naved001 (approved):

knikolla (approved):

PR #60, opened 2024-06-06
Loads invoice "rate" field as string to suppress scientific notation representation

### LINKED ISSUE #56: Output 0.000009 instead of 9.00E-06
Continuation of https://github.com/nerc-project/coldfront-plugin-cloud/issues/157

The representation issue might be because that column is read as a float. (No operation are performed with it so that's not really a problem)

```
>>> a = "0.000009"
>>> float(a)
9e-06
>>> Decimal(a)
Decimal('0.000009')
```

QuanMPhm:
Closes #56. Some of our rates may be have very low values, i.e 0.0000009, which may be displayed in scientific notation if read as a float by pandas.read_csv() Specifying the field type as str suppresses this behavior.


naved001 (approved):
This is okay as this is purely for appearance and we are not doing any math with it.

PR #61, opened 2024-06-11
Refactored billable invoice

### LINKED ISSUE #64: Refactor the billable invoice

QuanMPhm:
Closes #64, I have refactored the billable invoice. I'll explain in more detail below, and have also have a question that I would like the professionals' opinions on.

For some context, the billable invoice performs 2 preparation steps (removing nonbillables and validating PI names) and 3 distinct steps during processing (loading the old PI file, applying the new PI credit, and then writing back the old PI file). I decided to split these steps into their own functions, and to store those functions in `util.py`, instead of making them private functions in the`BillableInvoice` class.

As for the test cases, I created a new test class for the billable invoice. Note that this new test case uses randomly generated values, instead of a hardcoded dataframe. I figured this would be more robust?

While writing the test case, l also had a few design questions I wanted to ask...

Considering how unsightly some of the test cases are, especially for the [New PI credit](https://github.com/CCI-MOC/process_csv_report/blob/040055787597f03ec5ed180e6fbd376d8fd8bdcb/process_report/tests/unit_tests.py#L298), I was trying to think of ways to make the test case more organized and easier to follow. One of them was to create small test functions that would only test for one specific scenario (i.e a case for new PI, 1-month PI, overwritten entry, etc...), and to make a generator function that would that randomly generate the invoice and old-pi file (along with the expected output after they are processed) for a PI. I'm follow your opinions on whether randomized test data, instead of hardcoded data, would be better.

### DIFF
--- a/process_report/invoices/billable_invoice.py
+++ b/process_report/invoices/billable_invoice.py
@@ -16,41 +16,17 @@ logging.basicConfig(level=logging.INFO)
 
 @dataclass
 class BillableInvoice(invoice.Invoice):
+    NEW_PI_CREDIT_CODE = "0002"
+    INITIAL_CREDIT_AMOUNT = 1000
+    EXCLUDE_SU_TYPES = ["OpenShift GPUA100SXM4", "OpenStack GPUA100SXM4"]
+    PI_S3_FILEPATH = "PIs/PI.csv"
+
     nonbillable_pis: list[str]
     nonbillable_projects: list[str]
     old_pi_filepath: str
 
-    def _prepare(self):
-        self.data = self._remove_nonbillables(
-            self.data, self.nonbillable_pis, self.nonbillable_projects
-        )
-        self.data = self._validate_pi_names(self.data)
-
-    def _process(self):
-        old_pi_df = self._load_old_pis(self.old_pi_filepath)
-        self.data, updated_old_pi_df = self._apply_credits_new_pi(self.data, old_pi_df)
-        self._dump_old_pis(self.old_pi_filepath, updated_old_pi_df)
-
-    def _remove_nonbillables(
-        self,
-        data: pandas.DataFrame,
-        nonbillable_pis: list[str],
-        nonbillable_projects: list[str],
-    ):
-        return data[
-            ~data[invoice.PI_FIELD].isin(nonbillable_pis)
-            & ~data[invoice.PROJECT_FIELD].isin(nonbillable_projects)
-        ]
-
-    def _validate_pi_names(self, data: pandas.DataFrame):
-        invalid_pi_projects = data[pandas.isna(data[invoice.PI_FIELD])]
-        for i, row in invalid_pi_projects.iterrows():
-            logger.warn(
-                f"Billable project {row[invoice.PROJECT_FIELD]} has empty PI field"
-            )
-        return data[~pandas.isna(data[invoice.PI_FIELD])]
-
-    def _load_old_pis(self, old_pi_filepath) -> pandas.DataFrame:
+    @staticmethod
+    def _load_old_pis(old_pi_filepath) -> pandas.DataFrame:
         try:
             old_pi_df = pandas.read_csv(
                 old_pi_filepath,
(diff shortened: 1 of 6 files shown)


larsks:
I would like to see the second commit updated so that the commit message has some description of the changes involved in "Refactored billable invoice".

process_report/invoices/billable_invoice.py
```diff
@@ -0,0 +1,22 @@
...
+    def _prepare(self):
+        self.data = util.remove_nonbillables(
+            self.data, self.nonbillable_pis, self.nonbillable_projects
+        )
+        self.data = util.validate_pi_names(self.data)
+
+    def _process(self):
+        old_pi_df = util.load_old_pis(self.old_pi_filepath)
+        self.data, updated_old_pi_df = util.apply_credits_new_pi(self.data, old_pi_df)
+        util.dump_old_pis(self.old_pi_filepath, updated_old_pi_df)
```
naved001, line 22: I am guilty of doing this myself with code I've written but I think dumping these functions in a `utils` module may not be appropriate here. These functions are clearly preparing and processing the data here, you could create methods in this class or the base class if they could be used in other derived classes. `utils.py` is just too vague. 

`utils.py` is okay for functions like `get_timestamp()`, `compare_dates()` etc. 

@larsks @knikolla any opinions on this?

larsks: We talked about this on the call today, but I figured I would comment explicitly that I agree with @naved001's comment.

QuanMPhm: I will place these functions in the `BillableInvoice` class then. I was on the fence about where to place these functions before submitting the PR, since I thought they might be used by other invoices in the future, and because I thought I should keep the 'purity' of the Invoice classes to only contain the `_process`, `_prepare`, etc.

knikolla: > I will place these functions in the `BillableInvoice` class then. I was on the fence about where to place these functions before submitting the PR, since I thought they might be used by other invoices in the future, and because I thought I should keep the 'purity' of the Invoice classes to only contain the `_process`, `_prepare`, etc.

No need to worry about preserving the purity of classes. You can and should add as many functions as you need to a class.

process_report/invoices/billable_invoice.py
```diff
@@ -0,0 +1,186 @@
...
+        except FileNotFoundError:
+            sys.exit("Applying credit 0002 failed. Old PI file does not exist")
+
+        return old_pi_df
+
+    def _apply_credits_new_pi(
+        self, data: pandas.DataFrame, old_pi_df: pandas.DataFrame
+    ):
+        new_pi_credit_code = "0002"
+        INITIAL_CREDIT_AMOUNT = 1000
```
knikolla, line 74: These constants should be moved to class attributes with defaults.

process_report/invoices/billable_invoice.py
```diff
@@ -0,0 +1,186 @@
...
+        EXCLUDE_SU_TYPES = ["OpenShift GPUA100SXM4", "OpenStack GPUA100SXM4"]
+
+        data[invoice.CREDIT_FIELD] = None
```
knikolla, line 77: These 3 lines should be in the prepare method.

process_report/invoices/billable_invoice.py
```diff
@@ -0,0 +1,186 @@
...
+logging.basicConfig(level=logging.INFO)
+
+
+@dataclass
+class BillableInvoice(invoice.Invoice):
+    nonbillable_pis: list[str]
+    nonbillable_projects: list[str]
+    old_pi_filepath: str
+
+    def _prepare(self):
```
knikolla, line 23: Should this operate on a copy of the data? Or is a copy of the dataframe returned from `_remove_nonbillables`? I'm not very familiar with pandas so this is a bit of a basic question.

process_report/invoices/billable_invoice.py
```diff
@@ -0,0 +1,186 @@
...
+        self.data = self._remove_nonbillables(
+            self.data, self.nonbillable_pis, self.nonbillable_projects
+        )
+        self.data = self._validate_pi_names(self.data)
+
+    def _process(self):
+        old_pi_df = self._load_old_pis(self.old_pi_filepath)
```
knikolla, line 30: Move this to `_prepare` and assign it to `self.old_pi_df`.

process_report/invoices/billable_invoice.py
```diff
@@ -0,0 +1,186 @@
...
+                    credits_used != pi_old_pi_entry[credit_used_field]
+                ):
+                    print(
+                        f"Warning: PI file overwritten. PI {pi} previously used ${pi_old_pi_entry[credit_used_field]} of New PI credits, now uses ${credits_used}"
+                    )
+                old_pi_df.loc[
+                    old_pi_df[invoice.PI_PI_FIELD] == pi, credit_used_field
+                ] = credits_used
+
+        old_pi_df = old_pi_df.astype(
```
knikolla, line 154: Move this to `_prepare_export`.

knikolla (changes requested):
Some small comments to make better use of `_prepare` and `_prepare_export` methods. Implement them in a separate commit to preserve this diff as mostly a copy paste and dedicate more subtle refactoring changes to future commits (within this PR).

process_report/invoices/billable_invoice.py
QuanMPhm: When passed into a function, mutable objects like dataframes behave as if they're "passed-by-reference". Which is to say that unless I explicitly pass in a copy of the dataframe, every function could and will change the passed-in dataframe. 

I suppose this makes the return statements in most of the dataframe processing functions redundant. I can remove them at a later PR if you'd like.

process_report/invoices/billable_invoice.py
```diff
@@ -0,0 +1,186 @@
...
+    def _process(self):
+        old_pi_df = self._load_old_pis(self.old_pi_filepath)
+        self.data, updated_old_pi_df = self._apply_credits_new_pi(self.data, old_pi_df)
+        self._dump_old_pis(self.old_pi_filepath, updated_old_pi_df)
+
+    def _remove_nonbillables(
+        self,
+        data: pandas.DataFrame,
+        nonbillable_pis: list[str],
+        nonbillable_projects: list[str],
```
naved001, line 58: couldn't you access all these arguments with `self.data`, `self.nonbillable_pis`?

process_report/invoices/billable_invoice.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def _validate_pi_names(self, data: pandas.DataFrame):
+        invalid_pi_projects = data[pandas.isna(data[invoice.PI_FIELD])]
+        for i, row in invalid_pi_projects.iterrows():
+            logger.warn(
+                f"Billable project {row[invoice.PROJECT_FIELD]} has empty PI field"
+            )
+        return data[~pandas.isna(data[invoice.PI_FIELD])]
+
+    def _load_old_pis(self, old_pi_filepath) -> pandas.DataFrame:
```
naved001, line 73: you aren't using `self` in this method and you don't need it either, this should just be a staticmethod

naved001 (commented):
looks good, left a few questions/suggestions

process_report/invoices/billable_invoice.py
QuanMPhm: I suppose I could access the class attributes directly. I copied most of these invoice functions from `process_report.py` without changing the signature since I thought:
1. There will be less changes needed for the unit tests
2. I envisioned that only `_prepare`, `_process`, and `_prepare_process` would be the only functions that would reference their object attributes, with any other class function behaving as static functions rather than class functions. I had a vague feeling that this is right, and will follow your opinion.

process_report/invoices/billable_invoice.py
QuanMPhm: If they should be static methods, should they also be in `utils.py`?

process_report/invoices/billable_invoice.py
joachimweyl: @QuanMPhm if you moved these please resolve this conversation.

process_report/invoices/billable_invoice.py
joachimweyl: @QuanMPhm if you moved these 3 lines please resolve conversation.

process_report/invoices/billable_invoice.py
joachimweyl: @QuanMPhm if you are happy with your response please resolve the conversation.

process_report/invoices/billable_invoice.py
joachimweyl: @QuanMPhm if you are done moving this please resolve conversation.

process_report/invoices/billable_invoice.py
joachimweyl: @QuanMPhm if you moved this please resolve conversation.

process_report/invoices/billable_invoice.py
knikolla: > If they should be static methods, should they also be in `utils.py`?

No

QuanMPhm:
@naved001 @knikolla I've implemented all feedback given so far, and did a bit of refactoring with my most recent pushed-up changes.

process_report/process_report.py
```diff
@@ -202,7 +154,7 @@ def main():
     parser.add_argument(
         "--output-file",
         required=False,
-        default="filtered_output.csv",
+        default="filtered_output",
```
naved001, line 157: since we are at it, might as well give it a better default name like `billable_invoice`

process_report/invoices/billable_invoice.py
```diff
@@ -0,0 +1,202 @@
...
+            ].empty or pandas.isna(
+                new_pi_credit_amount := first_month_processed_pis[
+                    invoice.PI_INITIAL_CREDITS
+                ].iat[0]
+            ):
+                new_pi_credit_amount = default_initial_credit_amount
+
+            return new_pi_credit_amount
+
+        new_pi_credit_amount = get_initial_credit_amount(
```
naved001, line 115: why did you make `get_initial_credit_amount` a function? it's the only call here, might as well just do this in-line.

naved001 (commented):
I am mostly nitpicking, but I think any method where you passed `self` but didn't use it should just be a staticmethod.

process_report/invoices/billable_invoice.py
knikolla: I'm okay with it being a function as it better describes what that block of code is doing, even if it's called only from one place.

process_report/tests/unit_tests.py
```diff
@@ -473,9 +476,9 @@ def setUp(self):
             )
             .astype(
                 {
-                    "Initial Credits": pandas.ArrowDtype(pyarrow.decimal128(21, 2)),
-                    "1st Month Used": pandas.ArrowDtype(pyarrow.decimal128(21, 2)),
-                    "2nd Month Used": pandas.ArrowDtype(pyarrow.decimal128(21, 2)),
+                    "Initial Credits": object,
```
knikolla, line 479: Why is this an object now?

knikolla (changes requested):
Some comments, but otherwise looks pretty close.

process_report/invoices/billable_invoice.py
```diff
@@ -16,20 +16,62 @@
...
+        self.updated_old_pi_df = self.updated_old_pi_df.astype(
+            {
+                invoice.PI_INITIAL_CREDITS: pandas.ArrowDtype(
+                    pyarrow.decimal128(21, 2)
+                ),
+                invoice.PI_1ST_USED: pandas.ArrowDtype(pyarrow.decimal128(21, 2)),
+                invoice.PI_2ND_USED: pandas.ArrowDtype(pyarrow.decimal128(21, 2)),
+            },
+        )
+        self._dump_old_pis(self.old_pi_filepath, self.updated_old_pi_df)
```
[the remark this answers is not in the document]
QuanMPhm: In the previous code, the old PI file is uploaded to S3 [here](https://github.com/CCI-MOC/process_csv_report/blob/287afea5875b32bf25b8727b2790c54f14137328/process_report/process_report.py#L326). As for the billable invoice, that is [done](https://github.com/CCI-MOC/process_csv_report/blob/287afea5875b32bf25b8727b2790c54f14137328/process_report/process_report.py#L316-L319) in the same it statement

process_report/tests/unit_tests.py
QuanMPhm: The reason I made the change is very silly and was a oversight when refactoring the test cases. This boiled down to me no longer letting the test case [type cast the output old_pi_df](https://github.com/CCI-MOC/process_csv_report/blob/287afea5875b32bf25b8727b2790c54f14137328/process_report/tests/unit_tests.py#L560-L566). I have reverted this change, so the diff should be simpler now.

knikolla (approved):

larsks (approved):

PR #62, opened 2024-06-11
Allow a PI's institution name to be search with subdomains

### LINKED ISSUE #59: Find PI's institute by through the order of most qualified domain name
The emails of certain PIs contains multiple subdomains (i.e `pi@a.b.edu`). Currently, we check for a PI's institute by only matching their full email domain (`a.b.edu`) to the list of known domains in `institute_map.json`. For an institution with many subdomains, some of which may match to other institutes, but also many that matches to itself, we would need to add many fully qualified domain names (that share the same second-level, or third-level domain, etc) that would only match to that institution. 

Instead of only performing one match against the full domain name, we would like to match through each domain level, from the most qualified to the least. 

For example, given the pi email `pi@a.b.edu`, we should first match against `a.b.edu`, then `b.edu`.

QuanMPhm:
Closes #59. Now, for PIs whose emails contain multiple subdomains (i.e `a@b.c.edu`), each of their subdomains will be matched against institute_map.json, from the most qualified to least qualified subdomain.

In the example of `a@b.c.edu`, this means `b.c.edu` would be matched first, then `c.edu`, then `edu`.

I have removed the "Harvard University" entries from `institute_map.json` made redundant by this new feature.


process_report/tests/unit_tests.py
```diff
@@ -234,29 +234,19 @@ def test_get_pi_institution(self):
...
-        )
-        self.assertEqual(
-            process_report.get_institution_from_pi(
-                institute_map, "pi@northeastern.edu"
-            ),
-            "Northeastern University",
-        )
+        answers = {
+            "q@bu.edu": "Boston University",
+            "c@mclean.harvard.edu": "McLean Hospital",
```
joachimweyl, line 239: would it be helpful to add these items to the test as well?
"childrens.harvard.edu" : "Boston Children's Hospital",
    "meei.harvard.edu"      : "Massachusetts Eye & Ear",
    "dfci.harvard.edu"      : "Dana-Farber Cancer Institute",
    "bwh.harvard.edu"       : "Brigham and Women's Hospital",
    "bidmc.harvard.edu"     : "Beth Israel Deaconess Medical Center",

naved001 (approved):

process_report/tests/unit_tests.py
QuanMPhm: If we want to test more domains, I guess the question would be do we want to test for all the domains in `institute_map.json`, or just a few examples that validates the `get_institution_from_pi` function. @naved001 What is your opinion?

naved001: @QuanMPhm I think it's okay to test the additional domains that @joachimweyl suggested.

naved001 (approved):

PR #65, opened 2024-07-01
Fixed the two new domains

QuanMPhm:
The two new domains added were not written properly, such that PIs with the email domains dartmouth.edu or chemistry.harvard.edu would never be matched


joachimweyl:
Thank you!

joachimweyl (approved):

PR #71, opened 2024-07-15
Refactored BU Internal Invoice

### LINKED ISSUE #67: Refactor BU Internal invoice

QuanMPhm:
Closes #67. A new `bu_internal_invoice.py` file has been added. Note that the BU Internal invoice depends on the billable invoice. The test case for the BU Internal invoice has been modified appropriately.

### DIFF
--- a/process_report/invoices/billable_invoice.py
+++ b/process_report/invoices/billable_invoice.py
@@ -1,13 +1,12 @@
 from dataclasses import dataclass
-from decimal import Decimal
 import logging
 import sys
 
 import pandas
 import pyarrow
 
-import process_report.invoices.invoice as invoice
-import process_report.util as util
+from process_report.invoices import invoice, discount_invoice
+from process_report import util
 
 
 logger = logging.getLogger(__name__)
@@ -15,7 +14,7 @@ logging.basicConfig(level=logging.INFO)
 
 
 @dataclass
-class BillableInvoice(invoice.Invoice):
+class BillableInvoice(discount_invoice.DiscountInvoice):
     NEW_PI_CREDIT_CODE = "0002"
     INITIAL_CREDIT_AMOUNT = 1000
     EXCLUDE_SU_TYPES = ["OpenShift GPUA100SXM4", "OpenStack GPUA100SXM4"]
@@ -89,7 +88,7 @@ class BillableInvoice(invoice.Invoice):
         self.data = self._validate_pi_names(self.data)
         self.data[invoice.CREDIT_FIELD] = None
         self.data[invoice.CREDIT_CODE_FIELD] = None
-        self.data[invoice.BALANCE_FIELD] = Decimal(0)
+        self.data[invoice.BALANCE_FIELD] = self.data[invoice.COST_FIELD]
         self.old_pi_df = self._load_old_pis(self.old_pi_filepath)
 
     def _process(self):
@@ -143,14 +142,17 @@ class BillableInvoice(invoice.Invoice):
 
         current_pi_set = set(data[invoice.PI_FIELD])
         for pi in current_pi_set:
-            pi_projects = data[data[invoice.PI_FIELD] == pi]
+            credit_eligible_projects = data[
+                (data[invoice.PI_FIELD] == pi)
+                & ~(data[invoice.SU_TYPE_FIELD].isin(self.EXCLUDE_SU_TYPES))
+            ]
             pi_age = self._get_pi_age(old_pi_df, pi, self.invoice_month)
             pi_old_pi_entry = old_pi_df.loc[
                 old_pi_df[invoice.PI_PI_FIELD] == pi
             ].squeeze()
 
             if pi_age > 1:
-                for i, row in pi_projects.iterrows():
+                for i, row in credit_eligible_projects.iterrows():
                     data.at[i, invoice.BALANCE_FIELD] = row[invoice.COST_FIELD]
             else:
                 if pi_age == 0:
@@ -176,25 +178,16 @@ class BillableInvoice(invoice.Invoice):
                     )
                     credit_used_field = invoice.PI_2ND_USED
 
-                initial_credit = remaining_credit
-                for i, row in pi_projects.iterrows():
-                    if (
-                        remaining_credit == 0
-                        or row[invoice.SU_TYPE_FIELD] in self.EXCLUDE_SU_TYPES
-                    ):
-                        data.at[i, invoice.BALANCE_FIELD] = row[invoice.COST_FIELD]
-                    else:
-                        project_cost = row[invoice.COST_FIELD]
-                        applied_credit = min(project_cost, remaining_credit)
+                credits_used = self.apply_flat_discount(
+                    data,
+                    credit_eligible_projects,
+                    remaining_credit,
+                    invoice.CREDIT_FIELD,
+                    invoice.BALANCE_FIELD,
+                    invoice.CREDIT_CODE_FIELD,
+                    self.NEW_PI_CREDIT_CODE,
+                )
 
-                        data.at[i, invoice.CREDIT_FIELD] = applied_credit
-                        data.at[i, invoice.CREDIT_CODE_FIELD] = self.NEW_PI_CREDIT_CODE
-                        data.at[i, invoice.BALANCE_FIELD] = (
-                            row[invoice.COST_FIELD] - applied_credit
-                        )
-                        remaining_credit -= applied_credit
-
-                credits_used = initial_credit - remaining_credit
                 if (pi_old_pi_entry[credit_used_field] != 0) and (
                     credits_used != pi_old_pi_entry[credit_used_field]
                 ):
--- a/process_report/invoices/bu_internal_invoice.py
+++ b/process_report/invoices/bu_internal_invoice.py
@@ -2,10 +2,11 @@ from dataclasses import dataclass
 from decimal import Decimal
 
 import process_report.invoices.invoice as invoice
+import process_report.invoices.discount_invoice as discount_invoice
 
 
 @dataclass
-class BUInternalInvoice(invoice.Invoice):
+class BUInternalInvoice(discount_invoice.DiscountInvoice):
     subsidy_amount: int
 
     def _prepare(self):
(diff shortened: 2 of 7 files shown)


process_report/invoices/bu_internal_invoice.py
```diff
@@ -0,0 +1,68 @@
...
+        for project in project_list:
+            project_mask = self.data["Project"] == project
+            no_dup_project_mask = data_no_dup["Project"] == project
+
+            sum_fields_sums = self.data[project_mask][sum_fields].sum().values
+            data_no_dup.loc[no_dup_project_mask, sum_fields] = sum_fields_sums
+
+        self.data = self._apply_subsidy(data_no_dup, self.subsidy_amount)
+
+    def _apply_subsidy(self, dataframe, subsidy_amount):
```
knikolla, line 49: I think this is a great opportunity to try and reuse the logic from applying credits here.

knikolla (changes requested):

process_report/invoices/bu_internal_invoice.py
QuanMPhm: I have added a new function in `util.py` that contains the shared logic between the `apply_subsidy` and `apply_new_pi_credit` functions.

knikolla: @QuanMPhm I think a better location for that function would be a common parent class for both invoice classes.

knikolla (changes requested):

process_report/invoices/bu_internal_invoice.py
QuanMPhm: @knikolla Could I put the function in the base `Invoice`, since I think the Billable and BU Internal invoice have much "relatedness" to warrant making a new parent class just to implement them. I think applying a flat discount would be something any invoice might want to do

knikolla: I would prefer not. While other invoices might want to implement a flat discount, the subsidies and credits invoices ALWAYS do.

QuanMPhm:
@knikolla This may be too much but...since I'll be doing more nuanced refactoring with the BU Internal invoice, should that be a separate commit as well? Some of the "nuanced refactoring" has already been added to the first commit, so I wonder if its still worth the effort to split to commit and reorganize things.

knikolla:
> @knikolla This may be too much but...since I'll be doing more nuanced refactoring with the BU Internal invoice, should that be a separate commit as well? Some of the "nuanced refactoring" has already been added to the first commit, so I wonder if its still worth the effort to split to commit and reorganize things.

Think of it this way. Your goal is to make reviewing a PR as easy as possible. Splitting things into different commits often enables that by grouping a set of changes together with a clear commit message and description. This allows one commit to **do one thing and do it well and clearly**. 

To understand what making a PR easier to review means, you should go through the process of reviewing your own PRs (and obviously the code of others). Some questions to think about.

- Is it clear what the PR is trying to accomplish (and each commit)? (clear scope?)
- Is it hard to know where to even start reviewing this PR/commit? How could you make it easier for a reviewer to know where to start? (too big?)
- Once you start reviewing, how do you progress? Do you go top to bottom? Do you find yourself scrolling up and down? Which parts do you find yourself coming back to the most?
- Are there particular parts that are really hard for you to understand? How could you make those parts easier to review and understand?

Sometimes just splitting things into several commits significantly helps. Sometimes it doesn't. I think in this case it will. 

But ultimately, there is no single best answer on how to organize changes and you shouldn't seek to find an answer that fits every solution. Perfect is the enemy of done, and engineering software is a game of trade offs. You're trying to optimize both your time, the time of reviewers, your cognitive workload, and the cognitive workload of reviewers.

process_report/invoices/bu_internal_invoice.py
joachimweyl: @QuanMPhm what are next steps here?

QuanMPhm: I'm splitting up the commit for this PR to better show which commit copies over features, and which does more detailed refactoring.

QuanMPhm:
@knikolla I have re-organized the commits. Now, the first commit will only show code copying and the initial implementation of the BU-Internal invoice. The second one contains more detailed refactoring decisions.

process_report/invoices/discount_invoice.py
```diff
@@ -0,0 +1,79 @@
...
+        and applies a flat discount to those PI projects. Note that this function
+        will change the provided `invoice` Dataframe directly. Therefore, it does
+        not return the changed invoice.
+
+        This function assumes that the balance field shows the remaining cost of the project.
+        Basically, the provided invoice must show that balance = cost.
+
+        If the optional parameters `code_field` and `discount_code` are passed in,
+        `discount_code` will be comma-APPENDED to the `code_field` of projects where
+        the discount is applied
```
naved001, line 36: @joachimweyl wouldn't it be a problem to have command separated field in what will be ultimately be a comma separated report?

process_report/process_report.py
```diff
@@ -301,22 +253,40 @@ def main():
...
+    billable_inv.process()
+    billable_inv.export()
+    if args.upload_to_s3:
+        bucket = get_invoice_bucket()
+        billable_inv.export_s3(bucket)
+
+    bu_internal_inv = bu_internal_invoice.BUInternalInvoice(
+        name=args.BU_invoice_file,
+        invoice_month=invoice_month,
+        data=billable_inv.data,
```
naved001, line 276: will this modify the `billable_inv` dataframe? The reason I ask this is because we could consolidate all the s3 exports later. something like:

```python
if args.upload_to_s3:
    bucket = get_invoice_bucket()
    billable_inv.export_s3(bucket)
    bu_internal_inv.export_s3(bucket)
 ```

**Review by naved001 — COMMENTED** [review]
will continue to review this. Thanks for all the docstrings!

File: process_report/process_report.py (hunk 3 above)
Comment (naved001) [reply process_report/process_report.py:276]: also, it's just good to understand if this does modify it because then it would mean that after this runs "billable_inv" is no longer what it says (it may have turned into bu_internal_inv)

Comment (QuanMPhm) [reply process_report/process_report.py:276]: If I don't explicitly call `copy()` on a dataframe, then you can assume that the same dataframe is being assigned by reference:
```
a = pd.DataFrame({'a': [1,2,3,4],
                  'b': [1,2,3,4]})
b = a
b.loc[0, 'a'] = 100
print(a)
print(b)>>> 
>>> a = pd.DataFrame({'a': [1,2,3,4],
...                   'b': [1,2,3,4]})
>>> b = a
>>> b.loc[0, 'a'] = 100
>>> print(a)
     a  b
0  100  1
1    2  2
2    3  3
3    4  4
>>> print(b)
     a  b
0  100  1
1    2  2
2    3  3
3    4  4 
```

I will make sure each invoice's dataframes won't reference each other when exported together like in your suggestion.

Comment (QuanMPhm) [reply process_report/process_report.py:289]: @naved001  I'll consolidate all the `export()` and `export_s3()` calls together after your detailed review.

Comment (naved001) [reply process_report/process_report.py:289]: @QuanMPhm thanks, this largely looks fine to me.

Comment (QuanMPhm) [reply process_report/process_report.py]: While consolidating the export statements, I thought that it would make the code more organized if I process and export the invoices in the same code block. So instead of consolidating all the `export()` calls together, I defined [`process_and_export_invoices`](https://github.com/CCI-MOC/process_csv_report/pull/71/files#diff-fd9ba5f8a61085103fd99900c2be8be11438be8fb9919be0868c50b4a5df59b0R65-R71) to process and export a list of invoices. 

For now, `process_and_export_invoices` is called twice. First to export the nonbillable, billable, and lenovo invoice. Second to export the HU-BU and BU-Internal, since these two depends on processing that's done in the billable invoice.

File: process_report/invoices/billable_invoice.py [hunk 4]
```diff
@@ -1,12 +1,12 @@
 from dataclasses import dataclass
-from decimal import Decimal
 import logging
 import sys
 
 import pandas
 import pyarrow
 
 import process_report.invoices.invoice as invoice
+import process_report.invoices.discount_invoice as discount_invoice
```
knikolla, line 9: nitpick: it is rare to see `import a.b as b` because a more concise syntax is `from a import b`.

It avoids you having to repeat `b` twice, but it also allows you to specify two modules in one line. 

```
from process_report.invoices import invoice, discount_invoice
```

process_report/invoices/discount_invoice.py
```diff
@@ -0,0 +1,79 @@
...
+        discount_code: str = None,
+    ):
+        """
+        Takes in an invoice and a list of PI projects that are a subset of it,
+        and applies a flat discount to those PI projects. Note that this function
+        will change the provided `invoice` Dataframe directly. Therefore, it does
+        not return the changed invoice.
+
+        This function assumes that the balance field shows the remaining cost of the project.
+        Basically, the provided invoice must show that balance = cost.
```
knikolla, line 32: Balance field showing the remaining cost of a project is not the same as balance = cost. In the case of a previous invoice having applied a discount, balance would be different from cost, and that is ok. 

You bring a good point though with regards to Balance having to be preinitialized before calling apply flat discount.

process_report/invoices/billable_invoice.py
```diff
@@ -143,7 +143,10 @@ def get_initial_credit_amount(
 
         current_pi_set = set(data[invoice.PI_FIELD])
         for pi in current_pi_set:
-            pi_projects = data[data[invoice.PI_FIELD] == pi]
+            pi_projects = data[
+                (data[invoice.PI_FIELD] == pi)
+                & ~(data[invoice.SU_TYPE_FIELD].isin(self.EXCLUDE_SU_TYPES))
```
knikolla, line 148: You have changed the purpose of this variable, this `pi_projects` doesn't correctly express what this variable does anymore.

Rename it to something like `new_pi_credit_eligible_projects`

process_report/invoices/discount_invoice.py
```diff
@@ -0,0 +1,79 @@
...
+        :param pi_projects: A subset of `invoice`, containing all projects for a PI you want to apply the discount
+        :param discount_amount: The discount given to the PI
+        :param discount_field: Name of the field to put the discount amount applied to each project
+        :param balance_field: Name of the balance field
+        :param code_field: Name of the discount code field
+        :param discount_code: Code of the discount
+        """
+
+        def apply_discount_on_project(remaining_discount_amount, project_i, project):
+            remaining_project_cost = project[balance_field]
```
knikolla, line 50: This really is `remaining_project_balance` and should be named appropriately.

knikolla (changes requested):
Looks almost there!

process_report/invoices/discount_invoice.py
QuanMPhm: I should have clarified that the balance field should equal **what the PI would pay if the flat discount is not applied**, not just cost. I'll clarify this in the docstring.

knikolla (approved):

knikolla:
@QuanMPhm can you fix the merge conflicts? 

@naved001 would appreciate if you can you try to give this a review this week.

QuanMPhm:
@naved001  I have rebased the PR.

naved001 (approved):

PR #73, opened 2024-07-17
Refactored PI-specific invoices

### LINKED ISSUE #69: Refactor the PI-specific invoices

QuanMPhm:
Closes #69. The PI-specific is now implemented in the `PIInvoice` class. This class takes in the entire billable invoice and will export the PI invoices to local storage and S3. Because of its unique export file paths, it overrides the `export()`

Note that to follow the file path has changed slightly, from:
`f"{self.name}/{pi_instituition}_{pi}_{self.invoice_month}.csv"`
to...
`f"{self.name}/{pi_instituition}_{pi} {self.invoice_month}.csv"`

Note that replacement of the "_" to " ". I do this to be more consistent with the filename pattern for the other invoices.
Aside from that, I've also removed `upload_to_s3()` and its corresponding unit test, since it is no longer needed (after this PR, and certainly after all the refactor PRs are merged). @knikolla @naved001 Should I add a test for each invoice to see if they are exporting to desired S3 paths?

### DIFF
--- a/process_report/invoices/pi_specific_invoice.py
+++ b/process_report/invoices/pi_specific_invoice.py
@@ -1,0 +1,43 @@
+import os
+from dataclasses import dataclass
+
+import pandas
+
+import process_report.invoices.invoice as invoice
+import process_report.util as util
+
+
+@dataclass
+class PIInvoice(invoice.Invoice):
+    def _prepare(self):
+        self.pi_list = self.data[invoice.PI_FIELD].unique()
+
+    def export(self):
+        def _export_pi_invoice(pi):
+            if pandas.isna(pi):
+                return
+            pi_projects = self.data[self.data[invoice.PI_FIELD] == pi]
+            pi_instituition = pi_projects[invoice.INSTITUTION_FIELD].iat[0]
+            pi_projects.to_csv(
+                f"{self.name}/{pi_instituition}_{pi} {self.invoice_month}.csv"
+            )
+
+        if not os.path.exists(
+            self.name
+        ):  # self.name is name of folder storing invoices
+            os.mkdir(self.name)
+
+        for pi in self.pi_list:
+            _export_pi_invoice(pi)
+
+    def export_s3(self, s3_bucket):
+        def _export_s3_pi_invoice(pi_invoice):
+            pi_invoice_path = os.path.join(self.name, pi_invoice)
+            striped_invoice_path = os.path.splitext(pi_invoice_path)[0]
+            output_s3_path = f"Invoices/{self.invoice_month}/{striped_invoice_path}.csv"
+            output_s3_archive_path = f"Invoices/{self.invoice_month}/Archive/{striped_invoice_path} {util.get_iso8601_time()}.csv"
+            s3_bucket.upload_file(pi_invoice_path, output_s3_path)
+            s3_bucket.upload_file(pi_invoice_path, output_s3_archive_path)
+
+        for pi_invoice in os.listdir(self.name):
+            _export_s3_pi_invoice(pi_invoice)
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -1,5 +1,4 @@
 import argparse
-import os
 import sys
 import datetime
 
@@ -14,6 +13,7 @@ from process_report.invoices import (
     billable_invoice,
     NERC_total_invoice,
     bu_internal_invoice,
+    pi_specific_invoice,
 )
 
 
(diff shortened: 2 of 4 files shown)


process_report/invoices/pi_specific_invoice.py
```diff
@@ -0,0 +1,38 @@
...
+            self.name
+        ):  # self.name is name of folder storing invoices
+            os.mkdir(self.name)
+
+        pi_list = self.data[invoice.PI_FIELD].unique()
+
+        for pi in pi_list:
+            if pandas.isna(pi):
+                continue
+            pi_projects = self.data[self.data[invoice.PI_FIELD] == pi]
```
knikolla, line 24: This seems like a great candidate to be split into a separate method that takes a pi as an argument and exports that PI only.

process_report/invoices/pi_specific_invoice.py
```diff
@@ -0,0 +1,38 @@
...
+@dataclass
+class PIInvoice(invoice.Invoice):
+    def export(self):
+        self.local_to_s3_mapping = list()
+        if not os.path.exists(
+            self.name
+        ):  # self.name is name of folder storing invoices
+            os.mkdir(self.name)
+
+        pi_list = self.data[invoice.PI_FIELD].unique()
```
knikolla, line 19: This seems it should belong to `_prepare`, assigning a class attribute `self.pi_list`.

knikolla (changes requested):
A general note, try to split large sections of common code to separate clearly named functions or methods.

process_report/invoices/pi_specific_invoice.py
joachimweyl: @QuanMPhm how goes the splitting of this section?

QuanMPhm: It was resolved last week.

joachimweyl:
@knikolla please rereview.

knikolla (approved):

QuanMPhm:
@knikolla @naved001 Since all the invoices now implement their own S3 export functions, this meant that the `upload_to_s3` function in `process_report.py` was [removed](https://github.com/CCI-MOC/process_csv_report/pull/73/files#diff-162d25340846c53be6b9c357f242c269b4564a175d9f873c6a37a63239dd5419L486-L497), along with its [test case](https://github.com/CCI-MOC/process_csv_report/pull/73/files#diff-2ab8a00ce6228ed90b8929a4e45622b1586da07acbc85aa2d71533747f594ca9L790-L820) to check that files were uploaded to their intended s3 filepaths. I wanted to ask if this means I should implement a new test case that checks the export function of every invoice to do the same test.

knikolla:
> @knikolla @naved001 Since all the invoices now implement their own S3 export functions, this meant that the `upload_to_s3` function in `process_report.py` was [removed](https://github.com/CCI-MOC/process_csv_report/pull/73/files#diff-162d25340846c53be6b9c357f242c269b4564a175d9f873c6a37a63239dd5419L486-L497), along with its [test case](https://github.com/CCI-MOC/process_csv_report/pull/73/files#diff-2ab8a00ce6228ed90b8929a4e45622b1586da07acbc85aa2d71533747f594ca9L790-L820) to check that files were uploaded to their intended s3 filepaths. I wanted to ask if this means I should implement a new test case that checks the export function of every invoice to do the same test.

You can write a test for the exporting functionality of the base class `Invoice`, but it's unnecessary for invoices that don't modify that in any way.

QuanMPhm:
@knikolla @naved001 Since the `export_s3` function does not need to be concerned about file extensions anymore, I have simplified the current test case for s3 exporting. I also updated the commit message to reflect this change.

process_report/process_report.py
```diff
@@ -262,17 +262,13 @@ def main():
...
-
-    export_pi_billables(billable_inv.data.copy(), args.output_folder, invoice_month)
-
-    if args.upload_to_s3:
-        invoice_list = list()
-
-        for pi_invoice in os.listdir(args.output_folder):
-            invoice_list.append(os.path.join(args.output_folder, pi_invoice))
+    pi_inv = pi_specific_invoice.PIInvoice(
+        name=args.output_folder, invoice_month=invoice_month, data=billable_inv.data
```
naved001, line 266: so, for PI specific invoice you didn't pass in a copy of the billable_inv dataframe which is okay because we only export things for PIInvoice and no data is processed.

naved001 (approved):

process_report/process_report.py
QuanMPhm: Yes, that's correct

PR #98, opened 2024-09-25
Fix 3 domains in `institute_list.yaml`

QuanMPhm:
In #90, I did a poor job reviewing the domains being added.
@knikolla Should I add a unit test to validate institution domains in `institute_list.yaml` in some way?


knikolla (approved):
Looks good. No need for a test.

PR #105, opened 2024-09-25
Allow invoice subclasses to explicitly declare their exported columns and column names

### LINKED ISSUE #100: Allow invoices to explicitly declare their exported columns
As part of refactoring how invoice balances and processing is handled, each invoice class must declare a list of columns they are exporting.

QuanMPhm:
Closes #100. Two class attributes, `export_columns_list` and `exported_columns_map`, has been added to `Invoice`, along with a class function `_filter_columns()`. Subclasses of `Invoice` must now define `export_columns_list`, containing the ordered list of columns that must be exported in their respective invoices. Subclasses can optional define `exported_columns_map`, containing mappings between "internal" column names and what their name should be when exported.

The field name `RATE_FIELD` has been added to `invoice.py`. It was previously forgotten.

### DIFF
--- a/process_report/invoices/NERC_total_invoice.py
+++ b/process_report/invoices/NERC_total_invoice.py
@@ -12,6 +12,24 @@ class NERCTotalInvoice(invoice.Invoice):
         "University of Rhode Island",
     ]
 
+    export_columns_list = [
+        invoice.INVOICE_DATE_FIELD,
+        invoice.PROJECT_FIELD,
+        invoice.PROJECT_ID_FIELD,
+        invoice.PI_FIELD,
+        invoice.INVOICE_EMAIL_FIELD,
+        invoice.INVOICE_ADDRESS_FIELD,
+        invoice.INSTITUTION_FIELD,
+        invoice.INSTITUTION_ID_FIELD,
+        invoice.SU_HOURS_FIELD,
+        invoice.SU_TYPE_FIELD,
+        invoice.RATE_FIELD,
+        invoice.COST_FIELD,
+        invoice.CREDIT_FIELD,
+        invoice.CREDIT_CODE_FIELD,
+        invoice.BALANCE_FIELD,
+    ]
+
     @property
     def output_path(self) -> str:
         return f"NERC-{self.invoice_month}-Total-Invoice.csv"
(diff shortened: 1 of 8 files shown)


process_report/invoices/NERC_total_invoice.py
```diff
@@ -12,6 +12,24 @@ class NERCTotalInvoice(invoice.Invoice):
         "University of Rhode Island",
     ]
 
+    export_columns_list = [
```
naved001, line 15: should this just be in the base invoice class? Then sublcasses can override this when needed. It looks like NERCTotalInvoice, billable_invoice, and nonbillable_invoice all export the same columns.

knikolla: @naved001  I think I prefer an empty list. I wouldn't worry about the repetition for now as it helps us highlight similarities that we can later refactor into common abstractions.

process_report/invoices/invoice.py
```diff
@@ -78,7 +82,14 @@ def _prepare_export(self):
         that should or should not be exported after processing."""
         pass
 
+    def _filter_columns(self):
+        """Filters and renames columns before exporting"""
+        self.data = self.data[self.export_columns_list].rename(
```
knikolla, line 87: This should NOT rename the dataframe at `self.data.`.

knikolla (changes requested):

process_report/invoices/NERC_total_invoice.py
naved001: sure, sounds good.

process_report/invoices/invoice.py
QuanMPhm: From the Pandas documentation, `rename()` defaults to creating a copy of the dataframe. Though I have explicitly used `copy()` to make the intent clear.

knikolla: Regardless, the `self.data` attribute should point to the reference of original Dataframe that the Invoice was instantiated with. Do not reassign this `self.data`.

QuanMPhm: I have applied your suggestion, and also added the `export_columns_list` for the PI-specific invoice, which I forgot to add while rebasing.

PR #106, opened 2024-09-25
Implemented `Processor` class and refactored some preliminary processing 

### LINKED ISSUE #99: Implement a `Processor` class to refactor invoice processing
As the first step of refactoring how invoice balances and processing is handled, we want to initialize a `Processor` class which subclasses from Invoice, containing only the logic to process, or manipulate invoice data.

QuanMPhm:
Closes #99. Somewhat dependant on #100. A `Processor` class has been added, subclassing from the `Invoice` class. This is the first step to refactor invoicing in order to seperate the processing and filtering/exporting functionalities of our current `Invoice` subclasses. Subclasses of `Processor` should only process invoices and manipulate its internal data, while subclasses of `Invoice` should only perform filtering and exporting, never changing any data itself.

In addition to implementing `Processor`, two of its subclasses, `ValidatePIAliasProcessor` and `AddInstitutionProcessor` has been added to perform some preliminary processing steps.

### DIFF
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -15,7 +15,10 @@ from process_report.invoices import (
     bu_internal_invoice,
     pi_specific_invoice,
 )
-
+from process_report.processors import (
+    validate_pi_alias_processor,
+    add_institution_processor,
+)
 
 ### PI file field names
 PI_PI_FIELD = "PI"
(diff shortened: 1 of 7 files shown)


process_report/processors/validate_pi_alias_processor.py
```diff
@@ -0,0 +1,23 @@
...
+    def _validate_pi_aliases(dataframe: pandas.DataFrame, alias_dict: dict):
+        for pi, pi_aliases in alias_dict.items():
+            dataframe.loc[
+                dataframe[invoice.PI_FIELD].isin(pi_aliases), invoice.PI_FIELD
+            ] = pi
+
+        return dataframe
+
+    def _process(self):
+        self.data = self._validate_pi_aliases(self.data, self.alias_map)
```
naved001, line 23: 1. Why is `_validate_pi_aliases` a static method?
2. Because we directly modify the pandas dataframe, there's no reason to return or assign things. 

What if we wrote it like this:
```python
class ValidatePIAliasProcessor(processor.Processor):
    alias_map: dict

    def _validate_pi_aliases(self):
        for pi, pi_aliases in alias_map.items():
            self.data.loc[
                self.data[invoice.PI_FIELD].isin(pi_aliases), invoice.PI_FIELD
            ] = pi

    def _process(self):
        self._validate_pi_aliases()
```

Not that it matters much, but just curious as I have noticed this pattern in other places too.

QuanMPhm: I have changed this function, and another one in the `add_institute_processor.py` to use `self.data`, instead of having `data` passed in as an argument. This does lead to some inconsistency with other sections of code, specifically the invoice code. 

@knikolla @naved001  I will make a larger, sweeping commit to make this code design consistent if you think it's worth it.

naved001: @QuanMPhm eh, it's fine. let's deal with that later and we can focus on the important refactoring for now.

knikolla: ++, this refactoring is happening to enable an effective implementation of the prepaid groups work, everything more stylistic can wait until that is done.

process_report/process_report.py
```diff
@@ -220,15 +202,27 @@ def main():
...
+
     lenovo_inv = lenovo_invoice.LenovoInvoice(
-        name=args.Lenovo_file, invoice_month=invoice_month, data=merged_dataframe.copy()
+        name=args.Lenovo_file, invoice_month=invoice_month, data=add_institute_proc.data
     )
     nonbillable_inv = nonbillable_invoice.NonbillableInvoice(
         name=args.nonbillable_file,
         invoice_month=invoice_month,
-        data=merged_dataframe.copy(),
+        data=add_institute_proc.data,
```
naved001, line 225: same question

naved001 (approved):

knikolla (approved):

Issue #102, opened 2024-09-25
Refactor processing for the BU Subsidy


QuanMPhm:
Dependant on #99, the BU subsidy should be implemented as a `Processor` subclass.

Issue #101, opened 2024-09-25
Refactor processing for the New-PI credit


QuanMPhm:
Dependant on #108, the implementation of the New-PI credit should be implemented as a `Processor` subclass.

QuanMPhm:
@knikolla @naved001 While writing the PR for this issue, it occurred to me that the `New-PI Processor` will be dependant on the `ValidateBillablePIsProcessor` proposed in #108. This is because we would want to know which projects or PIs are non-billable before applying the New-PI discount.

As such, the `New-PI Processor` has a dependancy (`ValidateBillablePIsProcessor`), it also follows that `BillableInvoice` has a dependancy as well (`New-PI Processor`). 

For the purpose of this issue, this only means that I believe this issue is dependant on #108. For the purpose of code structure, I just wanted to highlight the fact that our `Invoice` and `Processor` classes may have dependencies of one (or more) `Processor`, in case this is something you guys might have concerns with.

PR #108, opened 2024-09-25
Implemented processors for removing nonbillables and validating billable PIs

### LINKED ISSUE #104: Refactor processing for removing nonbillables and PI name validation
Dependant on #99, the Billable invoice contains minor processing to remove nonbillable PIs, as well as validating PI names. Both of these processing steps can be implemented as a `Processor` subclass.

QuanMPhm:
Closes #104. Dependant on #100 and #99. The processing steps of removing nonbillables and validating billables in the Billable Invoice has been moved into `Processor` subclasses.

@knikolla @naved001 Because future processing steps (applying discounts, etc) assume that the invoice has gone through validation processing, this PR will be blocking other refactoring issues, namely #101 and #102.

### DIFF
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -17,6 +17,8 @@ from process_report.invoices import (
 from process_report.processors import (
     validate_pi_alias_processor,
     add_institution_processor,
+    remove_nonbillables_processor,
+    validate_billable_pi_processor,
 )
 
 ### PI file field names
@@ -227,15 +229,29 @@ def main():
         nonbillable_projects=projects,
     )
 
+    ### Remove nonbillables
+
+    remove_nonbillables_proc = remove_nonbillables_processor.RemoveNonbillables(
+        "", invoice_month, add_institute_proc.data, pi, projects
+    )
+    remove_nonbillables_proc.process()
+
+    validate_billable_pi_proc = (
+        validate_billable_pi_processor.ValidateBillablePIsProcessor(
+            "", invoice_month, remove_nonbillables_proc.data
+        )
+    )
+    validate_billable_pi_proc.process()
+
+    ### Initialize invoices
+
     if args.upload_to_s3:
         backup_to_s3_old_pi_file(old_pi_file)
 
     billable_inv = billable_invoice.BillableInvoice(
         name=args.output_file,
         invoice_month=invoice_month,
-        data=add_institute_proc.data.copy(),
-        nonbillable_pis=pi,
-        nonbillable_projects=projects,
+        data=validate_billable_pi_proc.data.copy(),
         old_pi_filepath=old_pi_file,
     )
 
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -2,7 +2,6 @@ import argparse
 import sys
 import datetime
 
-import json
 import pandas
 import pyarrow
 
@@ -15,7 +14,10 @@ from process_report.invoices import (
     bu_internal_invoice,
     pi_specific_invoice,
 )
-
+from process_report.processors import (
+    validate_pi_alias_processor,
+    add_institution_processor,
+)
 
 ### PI file field names
 PI_PI_FIELD = "PI"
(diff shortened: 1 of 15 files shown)


process_report/process_report.py
```diff
@@ -232,16 +234,32 @@ def main():
         nonbillable_projects=projects,
     )
 
+    ### Remove nonbillables
```
knikolla, line 237: Ideally, we want all Processor steps to happen before the Invoices steps. Is this not possible here?

QuanMPhm: With the way I'm writing the `RemoveNonbillables` Processor, the nonbillable projects are completely removed from the dataframe, so it would not be possible to perform this processing step before creating the Nonbillable-Invoice. A similar logic applies to `ValidateBillablePIsProcessor`.

I could have modified `RemoveNonbillables` so that instead of removing the nonbillable projects outright, it would add a column to "mark" which projects were non-billable. I just figured my current approach would result in the simplest diff.

Do you think I should still amend the PR so that all Processor steps happen before Invoice steps?

knikolla: @QuanMPhm I can see the advantage of having an `Is Billable?` (Y/N) column. Give it a shot and see how easily you can implement it. My concern is that a lot of the processing would then require checking for `Is Billable? = Y` before applying it, such as credits, etc.

I'll think about this a tiny bit more in the meanwhile. If you don't hear from me within Thursday ping me on Slack. Sometimes I lose track of these things.

QuanMPhm:
@knikolla I've implemented your feedback, and now all Processor steps are before the Invoices. @knikolla  @naved001 Let me know what you guys think. 

As a note to myself: I've stashed the earlier version of the PR in branch `104_draft`

process_report/processors/remove_nonbillables_processor.py
```diff
@@ -0,0 +1,27 @@
+from dataclasses import dataclass
+
+import pandas
+
+from process_report.invoices import invoice
+from process_report.processors import processor
+
+
+@dataclass
+class RemoveNonbillablesProcessor(processor.Processor):
```
knikolla, line 10: This name is now incorrect. Since there's not a lot happening here, I would suggest moving the logic of this processor inside of the validate billable pi processor.

knikolla (changes requested):
Now that you're not removing those rows from the dataframe where credits are applying, we do need to ignore non billable projects when applying credits, otherwise a PIs credits may be used from a project that is not billable.

knikolla:
> Now that you're not removing those rows from the dataframe where credits are applying, we do need to ignore non billable projects when applying credits, otherwise a PIs credits may be used from a project that is not billable.

Apologies. I see that you're doing this in the Billable invoice, since that hasn't be refactored yet. Disregard this comment.

QuanMPhm:
@knikolla @naved001 I have rebased the PR and addressed all current feedback

QuanMPhm:
Added `IS_BILLABLE_FIELD` and `MISSING_PI_FIELD` as constants in `invoice.py`

process_report/tests/util.py
```diff
@@ -80,3 +77,19 @@ def new_validate_pi_alias_processor(
...
 def new_lenovo_processor(name="", invoice_month="0000-00", data=pandas.DataFrame()):
     return lenovo_processor.LenovoProcessor(name, invoice_month, data)
+
+
+def new_validate_billable_pi_processor(
+    name="",
+    invoice_month="0000-00",
+    data=pandas.DataFrame(),
+    nonbillable_pis=[],
+    nonbillable_projects=[],
```
naved001, line 87: don't pass mutables as defaults, set it to `None` by default. This would apply to data, nonbillable_pis and nonbillable_projects

process_report/processors/validate_billable_pi_processor.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @staticmethod
+    def _get_billables(
+        data: pandas.DataFrame,
+        nonbillable_pis: list[str],
+        nonbillable_projects: list[str],
+    ):
+        return ~data[invoice.PI_FIELD].isin(nonbillable_pis) & ~data[
+            invoice.PROJECT_FIELD
+        ].isin(nonbillable_projects)
```
naved001, line 36: so this returns the boolean mask and not the data, right?

process_report/tests/util.py
QuanMPhm: After you told me, I realize now why this is a big problem. All of the functions in `tests/util.py` has this problem, so I believe it should be something I'll cover in a bug fix PR? Could I attach this problem to issue #111?

naved001: please do it as a separate issue and PR. However, this specific instance should be fixed it as part of this PR because it's introduced here

knikolla (commented):
Since nonbillables are not being removed from the dataframe now, are the processing steps where we are applying credits ignoring those nonbillable rows?

QuanMPhm:
@knikolla For now, the invoices that only care about billable projects (Billable, BU-Internal, etc) filter out billable projects like [this](https://github.com/CCI-MOC/process_csv_report/pull/108/files#diff-104a5729ddadcd79d71eaa18b7075216fd64bc29ead9190c7f8c172ca8ee20f4R82-R84)

knikolla (approved):
Ah, yes. I see it now. Looks good to me! 

Yes,  at the end of this set of refactorings we want to
1. Processors to operate on the same DataFrame (without doing `.copy`)
2. Invoices to filter out and shape the output how they want, without reassigning `self.data`.

naved001 (approved):

PR #112, opened 2024-11-07
Implemented processor for New-PI credit

### LINKED ISSUE #101: Refactor processing for the New-PI credit

QuanMPhm:
Closes #101. Depends on #108. I've implemented processor for applying the New-PI Credit, and included a new field to distinguish between the PI balance and the NERC balance. More details are in my commit message.

There are two commits to this PR, starting with `Initialized processor for applying the New-PI Credit and discounts`. The first one only copies over code, while the second commit actually makes functional changes.

### DIFF
--- a/process_report/invoices/NERC_total_invoice.py
+++ b/process_report/invoices/NERC_total_invoice.py
@@ -6,6 +6,12 @@ import process_report.util as util
 
 @dataclass
 class NERCTotalInvoice(invoice.Invoice):
+    """
+    This invoice operates on data processed by these Processors:
+    - ValidateBillablePIsProcessor
+    - NewPICreditProcessor
+    """
+
     INCLUDED_INSTITUTIONS = [
         "Harvard University",
         "Boston University",
@@ -45,6 +51,9 @@ class NERCTotalInvoice(invoice.Invoice):
         return f"Invoices/{self.invoice_month}/Archive/NERC-{self.invoice_month}-Total-Invoice {util.get_iso8601_time()}.csv"
 
     def _prepare_export(self):
+        self.data = self.data[
+            self.data[invoice.IS_BILLABLE_FIELD] & ~self.data[invoice.MISSING_PI_FIELD]
+        ]
         self.data = self.data[
             self.data[invoice.INSTITUTION_FIELD].isin(self.INCLUDED_INSTITUTIONS)
         ].copy()
--- a/process_report/processors/new_pi_credit_processor.py
+++ b/process_report/processors/new_pi_credit_processor.py
@@ -1,0 +1,174 @@
+import sys
+
+from dataclasses import dataclass
+import pandas
+import pyarrow
+
+from process_report import util
+from process_report.invoices import invoice
+from process_report.processors import discount_processor
+
+
+@dataclass
+class NewPICreditProcessor(discount_processor.DiscountProcessor):
+    NEW_PI_CREDIT_CODE = "0002"
+    INITIAL_CREDIT_AMOUNT = 1000
+    EXCLUDE_SU_TYPES = ["OpenShift GPUA100SXM4", "OpenStack GPUA100SXM4"]
+
+    old_pi_filepath: str
+    limit_new_pi_credit_to_partners: bool = False
+
+    @staticmethod
+    def _load_old_pis(old_pi_filepath) -> pandas.DataFrame:
+        try:
+            old_pi_df = pandas.read_csv(
+                old_pi_filepath,
+                dtype={
+                    invoice.PI_INITIAL_CREDITS: pandas.ArrowDtype(
+                        pyarrow.decimal128(21, 2)
+                    ),
+                    invoice.PI_1ST_USED: pandas.ArrowDtype(pyarrow.decimal128(21, 2)),
+                    invoice.PI_2ND_USED: pandas.ArrowDtype(pyarrow.decimal128(21, 2)),
+                },
+            )
+        except FileNotFoundError:
+            sys.exit("Applying credit 0002 failed. Old PI file does not exist")
+
+        return old_pi_df
+
+    @staticmethod
+    def _get_pi_age(old_pi_df: pandas.DataFrame, pi, invoice_month):
+        """Returns time difference between current invoice month and PI's first invoice month
+        I.e 0 for new PIs
+        Will raise an error if the PI'a age is negative, which suggests a faulty invoice, or a program bug"""
+        first_invoice_month = old_pi_df.loc[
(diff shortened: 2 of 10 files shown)


QuanMPhm:
I fixed a bug with `apply_discount_on_project()`. In its old version, the function could lead to a negative PI balance. 

@knikolla @naved001 I could write the test cases that would catch this mistake in this PR, but since the PR for #102 would catch this, I believe I will just write these tests in that PR?

naved001:
> I fixed a bug with apply_discount_on_project(). In its old version, the function could lead to a negative PI balance.

Could you say more about the impact of this bug? Would we need to rerun the invoice with the bug fix? 

> I could write the test cases that would catch this mistake in this PR, but since the PR for https://github.com/CCI-MOC/process_csv_report/issues/102 would catch this, I believe I will just write these tests in that PR?

Which commit has the bug fix? the commit message in the 3rd commit covers a bunch of things and I didn't find a note about the bug fix. Also, if the fix for a bug is in this PR I think the test should also be a part of this.

QuanMPhm:
@naved001 The bug fix was made when I amended my PR and force-pushed on Nov 11. [This](https://github.com/CCI-MOC/process_csv_report/compare/029985179d059e995a50ddb7e03757a0cfab623a..59c0012673a47f76724a330dd09df1a435328937) should show you the changes I made to fix the bug.

process_report/invoices/NERC_total_invoice.py
```diff
@@ -6,6 +6,12 @@
 
 @dataclass
 class NERCTotalInvoice(invoice.Invoice):
+    """
+    Dependancies:
+    - ValidateBillablePIsProcessor
+    - NewPICreditProcessor
+    """
```
larsks, line 13: What are these dependencies? Are we documenting them here (and elsewhere in this file) because they are otherwise non-obvious?

process_report/processors/new_pi_credit_processor.py
```diff
@@ -0,0 +1,190 @@
...
+from dataclasses import dataclass
+import pandas
+import pyarrow
+
+from process_report import util
+from process_report.invoices import invoice
+from process_report.processors import discount_processor
+
+
+@dataclass
```
larsks, line 12: Why is this a dataclass? It doesn't appear to have any class attributes. I guess the same question applies to `discount_process.DiscountProcessor` as well.

larsks (approved):

process_report/invoices/NERC_total_invoice.py
QuanMPhm: Many of the invoice subclasses expect certain columns to exist in the dataframe they are given, and some of those columns are initialized and populated by certain processors. `ValidateBillablePIsProcessor` populates the `IS_BILLABLE_FIELD` and `MISSING_PI_FIELD`, while `NewPICreditProcessor` populates many columns, such as `BALANCE_FIELD`. These columns are used by many invoice classes, including `NERCTotalInvoice`, during their filtering or exporting steps. As such, I believed that there are dependencies between invoices and processors, since many invoices need certain processors to run before them to work properly.

I thought that I should highlight these dependencies with these docstrings. I will follow your advice if you believe something better should be done.

process_report/invoices/NERC_total_invoice.py
```diff
@@ -45,6 +51,9 @@ def output_s3_archive_key(self):
         return f"Invoices/{self.invoice_month}/Archive/NERC-{self.invoice_month}-Total-Invoice {util.get_iso8601_time()}.csv"
 
     def _prepare_export(self):
+        self.data = self.data[
+            self.data[invoice.IS_BILLABLE_FIELD] & ~self.data[invoice.MISSING_PI_FIELD]
+        ]
         self.data = self.data[
```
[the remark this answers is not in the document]
QuanMPhm: This statement filters out rows from `self.data` where `IS_BILLABLE_FIELD` is False and `MISSING_PI_FIELD` is True. 

`self.data[invoice.IS_BILLABLE_FIELD] & ~self.data[invoice.MISSING_PI_FIELD]` would indeed return a boolean series.
`self.data[
            self.data[invoice.IS_BILLABLE_FIELD] & ~self.data[invoice.MISSING_PI_FIELD]
        ]` would return a view of `self.data` with rows filtered out.

process_report/invoices/NERC_total_invoice.py
larsks: I think it's fine to document this in the class docstring, but I think you should be a little bit more verbose. E.g., `This class requires the following columns to be defined in the input dataframe:`. This makes it clear that you're not referring to Python modules or classes or something else that could be considered a "dependency". If the required fields are only provided by a specific processor, it's probably better to say something like `This class operates on data that has been processed by NewPICreditProcess`.

Also note that the spelling is "dependencies".

process_report/invoices/NERC_total_invoice.py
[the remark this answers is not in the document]
larsks: I think the answer here, based on your comment, is that "`self.data` is not a bitfield, it's a dataframe, and these are pandas vector operations". Is that accurate?

process_report/processors/new_pi_credit_processor.py
QuanMPhm: `NewPICreditProcessor` does have `old_pi_filepath` and `limit_new_pi_credit_to_partners` as its class attributes, but as for `DiscountProcessor`, you are right, it has none. I guess I made them dataclasses for uniformity with all the other processors and invoices and didn't put too much thought into it. 

I will remove `@dataclass` from `DiscountProcessor`

process_report/invoices/NERC_total_invoice.py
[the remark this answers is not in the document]
QuanMPhm: Ah yes. Sorry for not clarifying

hakasapl (approved):

naved001 (approved):
This looks fine to me. However, I'd like in the future that PRs this size be broken down into smaller commits. 

It would also be nice if we split the unit tests, right now it's all one giant file.

QuanMPhm:
@naved001 My apologies. I realized at some point that the PR could have been split into another commit to include the new `PI_BALANCE`. Sorry for making you review the whole thing. 

I have made an issue to split up the unit test file [here](https://github.com/CCI-MOC/process_csv_report/issues/116)

naved001:
@knikolla do you think you'd have time to review this?

knikolla:
With 3 approvals I feel fine with this merging without my review.

PR #113, opened 2024-11-12
Implemented processor for BU Subsidy

### LINKED ISSUE #102: Refactor processing for the BU Subsidy

QuanMPhm:
Depends on #112. Closes #102. This PR consists of 2 commits, starting from `Initialized processor for BU subsidy...`

The first commit copies over code without changes.
The second commit makes functional changes.
Details about the commits are found in the commit messages.


process_report/tests/unit_tests.py
```diff
@@ -757,119 +757,179 @@ def test_apply_credit_error(self):
...
+        project_names=None,
+        institution=None,
+        is_billable=None,
+        missing_pi=None,
+    ):
+        if not balances:
+            balances = pi_balances
+
+        if not project_names:
+            project_names = ["" for _ in range(len(pi))]
```
naved001, line 795: If you don't care about the project name, I'd still give it a name so it's clear that it's not something you missed.

btw, this is also valid syntax if you just want repeat items in a list
```
>>> ['some-project']*len(pi)
['some-project', 'some-project', 'some-project', 'some-project']

```

naved001: @QuanMPhm how we initialize the list was merely a suggestion, feel free to ignore it. However, I do like giving things some names instead of an empty string `""`.

larsks: In the future, note that you can write instead:

```
project_names = [""] * len(pi)
```

No loops required!

larsks (approved):

PR #114, opened 2024-11-13
Removed mutable defaults from function signatures in `tests/util.py`

QuanMPhm:
Addressing Naved's comment [here](https://github.com/CCI-MOC/process_csv_report/pull/108#discussion_r1837089012).


naved001 (approved):
Thanks Quan! I am glad that all of our tests are still good!

PR #119, opened 2024-12-04
Invoices no longer assign to `self.data` during filtering step

### LINKED ISSUE #111: Invoice classes should not modify or reassign `self.data`
Currently, all `Invoice` subclasses reassign `self.data` before exporting their internal dataframes. This reassignment is done mostly because each invoice only needs to export certain columns. While this step does not change the internal dataframe's data, and has no unwanted side effects, it's presence still raises unwanted ambiguity, and goes against our belief that the `Invoice` subclasses should not modify their given dataframes.

QuanMPhm:
Closes #111. All invoice classes will now assign to `self.export_data`

This removes ambiguity on whether the invoices will modify their internal dataframe and the need to call `copy()` on the processed dataframe


naved001 (commented):
This looks fine to me but it broke your tests.

QuanMPhm:
I have fixed the PR

knikolla (approved):

PR #129, opened 2024-12-31
PI invoices are now exported as PDFs

### LINKED ISSUE #84: Generate PDFs from the PI-specific invoices
To help automate billing, in addition to exporting the PI-specific CSVs, the processing script should also generate PDFs for them. Based on the PDFs we are currently sending out to PIs, the only additional steps needed to turn the PI-specific CSVs into PDFs are:
- Appending of a row summing up the following numbers:
    - Cost
    - Credit 
    - Balance
- Naming convention for the PDFs are: "<Institution name>_<PI email>_<invoice month>.pdf"

Examples can be found in our billing PDFs.

Based on our prepaid scaffolding document, it seems we may want to update the naming for these PI-specific PDFs after implementing prepaid group billing.

@knikolla @naved001 @joachimweyl I believe this can be solved fairly simply with any python PDF-exporting package. Let me know if there is anything else I should be aware of before starting work on this issue.

QuanMPhm:
Closes #84. This PR consists of the last commit.
This PR mostly involved changes to the pi-specific invoice class. More details in the commit message. Two Python packages were crucial for this task, [Jinja](https://jinja.palletsprojects.com/en/stable/) and [Selenium](https://www.selenium.dev/documentation/) 

@knikolla @larsks I am aware that running the invoice script with this change will make pandas print a bunch of `FutureWarnings` for each PDF printed. Should I address them after the current state of the PR is approved?

### DIFF
--- a/process_report/invoices/pi_specific_invoice.py
+++ b/process_report/invoices/pi_specific_invoice.py
@@ -1,10 +1,22 @@
 import os
+import sys
 from dataclasses import dataclass
+import subprocess
+import tempfile
+import logging
 
 import pandas
+from jinja2 import Environment, FileSystemLoader
 
 import process_report.invoices.invoice as invoice
 import process_report.util as util
+
+
+TEMPLATE_DIR_PATH = "process_report/templates"
+
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
 
 
 @dataclass
@@ -14,6 +26,21 @@ class PIInvoice(invoice.Invoice):
     - ValidateBillablePIsProcessor
     - NewPICreditProcessor
     """
+
+    TOTAL_COLUMN_LIST = [
+        invoice.COST_FIELD,
+        invoice.CREDIT_FIELD,
+        invoice.BALANCE_FIELD,
+    ]
+
+    DOLLAR_COLUMN_LIST = [
+        invoice.RATE_FIELD,
+        invoice.GROUP_BALANCE_FIELD,
+        invoice.COST_FIELD,
+        invoice.GROUP_BALANCE_USED_FIELD,
+        invoice.CREDIT_FIELD,
+        invoice.BALANCE_FIELD,
+    ]
 
     export_columns_list = [
         invoice.INVOICE_DATE_FIELD,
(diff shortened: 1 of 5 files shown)


larsks (changes requested):
I think using `selenium` is overkill here. You can drive chrome's print-to-pdf functionality directly from the command line:

```
google-chrome --headless --print-to-pdf=output.pdf --no-pdf-header-footer document.html
```

This takes a single `subprocess.run()` call.

process_report/invoices/pi_specific_invoice.py
```diff
@@ -37,30 +61,134 @@ class PIInvoice(invoice.Invoice):
...
-            _export_pi_invoice(pi)
+            if pandas.isna(pi):
+                continue
+
+            pi_dataframe = self._get_pi_dataframe(export_data, pi)
+            pi_instituition = pi_dataframe[invoice.INSTITUTION_FIELD].iat[0]
+
+            invoice_html_path = _create_html_invoice()
+            _create_pdf_invoice()
+            os.remove(invoice_html_path)
```
larsks, line 191: Instead of manually cleaning up the temporary file, you could make more effective use of `tempfile.NamedTemporaryFile` here by passing an open file into your `_create_*` methods:

```
with tempfile.NamedTemporaryFile(mode='w', suffix='.html') as tempfd:
  _create_html_invoice(tempfd)
  _create_pdf_invoice(tempfd)
```

```
def _create_html_invoice(tempfd):
    environment = Environment(loader=FileSystemLoader(TEMPLATE_DIR_PATH))
    template = environment.get_template("pi_invoice.html")
    content = template.render(
        data=pi_dataframe,
    )

    tempfd.write(content)

    # You might need this to ensure the content is visible in the file.
    tempfd.flush()
```

And you would need similar changes in `_create_pdf_invoice`.

QuanMPhm: @larsks  It turns out I did need to flush it. Thank you!

QuanMPhm:
@larsks I have removed the `selenium` dependancy and followed your approach. However, know I have a question regarding the unit test...

@larsks @knikolla The unit test failed because it did not find the chromium binary. I should change the github action file to install Chrome, or just I mock the `subprocess.run()` call to just write a pdf file?

knikolla:
> @larsks @knikolla The unit test failed because it did not find the chromium binary. Should I change the github action file to install Chrome, or should I just mock the `subprocess.run()` call to just write a pdf file?

Mock `os.path.exists` and `subprocess.run` and verify that it was called with the parameters you expected it to be called with.

larsks:
> @larsks @knikolla The unit test failed because it did not find the chromium binary. Should I change the github action file to install Chrome, or should I just mock the `subprocess.run()` call to just write a pdf file?

*Unit* tests shouldn't have external dependencies, which means you should mock out `subprocess.run`. 

It might be nice to have an integration test that calls out to chrome; this would ensure that our chrome command line is valid. If @knikolla agrees that's a good idea, that could be a future pull request.

knikolla:
> > @larsks @knikolla The unit test failed because it did not find the chromium binary. Should I change the github action file to install Chrome, or should I just mock the `subprocess.run()` call to just write a pdf file?
> 
> _Unit_ tests shouldn't have external dependencies, which means you should mock out `subprocess.run`.
> 
> It might be nice to have an integration test that calls out to chrome; this would ensure that our chrome command line is valid. If @knikolla agrees that's a good idea, that could be a future pull request.

Considering the rapid pace of chrome development and inability to rely on semantic versioning as they bump the major version number too often, I think it's worth having a test in the future that verifies that the command line arguments that we're passing are still accepted by the current version of chrome.

For the purposes of this PR, a unit test is enough.

QuanMPhm:
I have created a new issue to create the integration test at #145. I have also implemented all feedback so far.

process_report/invoices/pi_specific_invoice.py
```diff
@@ -43,31 +70,109 @@ def _prepare(self):
         ]
         self.pi_list = self.export_data[invoice.PI_FIELD].unique()
 
+    def _get_pi_dataframe(self, data, pi):
+        def add_dollar_sign(data):
+            if pandas.isna(data):
+                return data
+            else:
+                return "$" + str(data)
```
larsks, line 78: You don't release need the `else` clause here, and prefer f-strings over string concatenation:

```suggestion
            return f"${data}"
```

But see my comment later on about whether or not we even need this function.

process_report/invoices/pi_specific_invoice.py
```diff
@@ -43,31 +70,109 @@ def _prepare(self):
...
+                    invoice.GROUP_INSTITUTION_FIELD,
+                    invoice.GROUP_BALANCE_FIELD,
+                    invoice.GROUP_BALANCE_USED_FIELD,
+                ],
+                axis=1,
+            )
+
+        # Add a row containing sums for certain columns
+        column_sums = list()
+        sum_columns_list = list()
```
larsks, line 96: In general, prefer `[]` over `list()` to initialize list variables:

```suggestion
        column_sums = []
        sum_columns_list = []
```

process_report/invoices/pi_specific_invoice.py
```diff
@@ -43,31 +70,109 @@ def _prepare(self):
...
+            template = environment.get_template("pi_invoice.html")
+            content = template.render(
+                data=pi_dataframe,
+            )
+            temp_fd.write(content)
+            temp_fd.flush()
+
+        def _create_pdf_invoice(temp_fd_name):
+            chrome_binary_location = os.environ.get(
+                "CHROME_BIN_PATH", "usr/bin/chromium"
```
larsks, line 130: Should this actually be `/usr/bin/chromium`?

```suggestion
                "CHROME_BIN_PATH", "/usr/bin/chromium"
```

process_report/invoices/pi_specific_invoice.py
```diff
@@ -43,31 +70,109 @@ def _prepare(self):
...
+                f"{self.name}/{pi_instituition}_{pi}_{self.invoice_month}.pdf"
+            )
+            subprocess.run(
+                [
+                    chrome_binary_location,
+                    "--headless",
+                    "--no-sandbox",
+                    f"--print-to-pdf={invoice_pdf_path}",
+                    "--no-pdf-header-footer",
+                    "file://" + temp_fd_name,
```
larsks, line 147: Prefer an f-string (especially since you're using this syntax just a few lines earlier):

```suggestion
                    f"file://{temp_fd_name}",
```

process_report/invoices/pi_specific_invoice.py
```diff
@@ -43,31 +70,109 @@ def _prepare(self):
...
+                ],
+                capture_output=True,
             )
 
         self._filter_columns()
+
         if not os.path.exists(
             self.name
         ):  # self.name is name of folder storing invoices
             os.mkdir(self.name)
```
larsks, line 157: You can simplify this using `os.makedirs`:

```suggestion
        os.makedirs(self.name, exist_ok=True)
```

process_report/invoices/pi_specific_invoice.py
```diff
@@ -43,31 +70,109 @@ def _prepare(self):
...
+        ] = None  # Adds a new row to end of dataframe initialized with None
+        pi_projects.loc[pi_projects.index[-1], invoice.INVOICE_DATE_FIELD] = "Total"
+        pi_projects.loc[pi_projects.index[-1], sum_columns_list] = column_sums
+
+        # Add dollar sign to certain columns
+        for column_name in self.DOLLAR_COLUMN_LIST:
+            if column_name in pi_projects.columns:
+                pi_projects[column_name] = pi_projects[column_name].apply(
+                    add_dollar_sign
+                )
```
larsks, line 112: The `add_dollar_sign` function is so trivial, and we're only calling it once, so it might be better to replace it with a lambda function:

```suggestion
                pi_projects[column_name] = pi_projects[column_name].apply(
                    lambda data: data if pandas.isna(data) else f"${data}"
                )
```

process_report/tests/unit/invoices/test_pi_specific_invoice.py
```diff
@@ -1,68 +1,147 @@
...
-        self.assertEqual(
-            len(os.listdir(output_dir.name)),
-            len(self.dataframe["Manager (PI)"].unique()),
+
+    def test_get_pi_dataframe(self):
+        def add_dollar_sign(data):
+            if pandas.isna(data):
+                return data
+            else:
+                return "$" + str(data)
```
larsks, line 46: (See my earlier comment about this function.)

process_report/tests/unit/invoices/test_pi_specific_invoice.py
```diff
@@ -1,68 +1,147 @@
...
+            chrome_arglist, _ = mock_subprocess_run.call_args_list[i]
+            answer_arglist = [
+                "usr/bin/chromium",
+                "--headless",
+                "--no-sandbox",
+                f"--print-to-pdf={pi_pdf_path}",
+                "--no-pdf-header-footer",
+            ]
+            for answer_arg in answer_arglist:
+                self.assertTrue(answer_arg in chrome_arglist[0])
```
larsks, line 147: I would be more constrained in your check:

```suggestion
            self.assertTrue(answer_arglist == chrome_arglist[0][:-1])
```

process_report/tests/unit/invoices/test_pi_specific_invoice.py
```diff
@@ -1,68 +1,147 @@
...
+            [100, 200, 300, 400],
+            group_name=[None, "G1", None, None],
         )
 
-        self.assertIn("ProjectA", pi_df["Project - Allocation"].tolist())
-        self.assertIn("ProjectB", pi_df["Project - Allocation"].tolist())
-        self.assertIn("ProjectC", pi_df["Project - Allocation"].tolist())
+        mock_filter_cols.return_value = test_invoice
+        mock_path_exists.return_value = True
+        output_dir = tempfile.TemporaryDirectory()
```
larsks, line 127: We don't need to create a temporary directory -- because we're mocking out `subprocess.run`, we never create any output.

You can do this instead:

```
        pi_inv = test_utils.new_pi_specific_invoice(
            "/fakedir", invoice_month, data=test_invoice
        )
        pi_inv.process()
        pi_inv.export()
        pi_pdf_1 = f"/fakedir/BU_PI1_{invoice_month}.pdf"
        pi_pdf_2 = f"/fakedir/HU_PI2_{invoice_month}.pdf"
```

process_report/tests/unit/invoices/test_pi_specific_invoice.py
```diff
@@ -1,68 +1,147 @@
...
-    def test_export_pi(self, mock_filter_cols):
-        mock_filter_cols.return_value = self.dataframe
+        if not missing_pi:
+            missing_pi = [False for _ in range(len(pi))]
 
-        output_dir = tempfile.TemporaryDirectory()
-        pi_inv = test_utils.new_pi_specific_invoice(
-            output_dir.name, invoice_month=self.invoice_month, data=self.dataframe
+        if not group_name:
+            group_name = [None for _ in range(len(pi))]
```
larsks, line 25: You can write instead:

```suggestion
            group_name = [None] * len(pi)
```

Compare:

```
>>> pi=[1,2,3,4,5]
>>> [None for _ in range(len(pi))]
[None, None, None, None, None]
>>> [None] * len(pi)
[None, None, None, None, None]
```

process_report/invoices/pi_specific_invoice.py
```diff
@@ -43,31 +70,109 @@ def _prepare(self):
...
+            temp_fd.write(content)
+            temp_fd.flush()
+
+        def _create_pdf_invoice(temp_fd_name):
+            chrome_binary_location = os.environ.get(
+                "CHROME_BIN_PATH", "usr/bin/chromium"
+            )
+            if not os.path.exists(chrome_binary_location):
+                sys.exit(
+                    f"Chrome binary does not exist at {chrome_binary_location}. Make sure the env var CHROME_BIN_PATH is set correctly or that Google Chrome is installed"
```
larsks, line 134: The message directs the user to ensure that Google Chrome is installed, but the code defaults to Chromium (which could lead to the situation in which Google Chrome is installed but the user still receives this error message). It makes more sense to replace `or` with `and`:

```suggestion
                    f"Chrome binary does not exist at {chrome_binary_location}. Make sure that Google Chrome is installed and the environment variable CHROME_BIN_PATH is set correctly."
```

joachimweyl:
@QuanMPhm is this ready to merge?

naved001 (approved):

PR #143, opened 2025-01-17
Implemented an invoice to export prepay groups invoices as PDFs

### LINKED ISSUE #137: Generate PDFs for Prepaid Groups
As part of the requirements listed in the prepayment scaffolding, we want to generate PDF invoices for prepay groups.

QuanMPhm:
Closes #137. This PR consists of the last commit. I feel like there is still a bit of polishing left to do, so I will submit this as a draft for now.

The PR is quite similar to #137 in terms of logic, using the chromium library to generate the PDFs.


process_report/invoices/MOCA_group_specific_invoice.py
```diff
@@ -0,0 +1,201 @@
...
+                [
+                    chrome_binary_location,
+                    "--headless",
+                    "--no-sandbox",
+                    f"--print-to-pdf={invoice_pdf_path}",
+                    "--no-pdf-header-footer",
+                    "file://" + temp_fd_name,
+                ],
+                capture_output=True,
+            )
```
larsks, line 175: This looks like a duplicate of the code in `process_report/invoices/pi_specific_invoice.py`. Can we factor this out into a common base class (or mixin)?

larsks (changes requested):

process_report/invoices/MOCA_group_specific_invoice.py
QuanMPhm: Because I believe #137 will undergo a lot of changes during review I will factor this into a base class after #137 is merged.

process_report/process_report.py
```diff
@@ -184,6 +185,12 @@ def main():
         default="pi_invoices",
         help="Name of output folder containing pi-specific invoice csvs",
     )
+    parser.add_argument(
+        "--prepay-groups-output-folder",
```
knikolla, line 189: Let's not make this configurable but provide default locations for invoices moving forward. Name should be hardcoded in the invoice.

knikolla (commented):
This PR is a good candidate for us to go over in a Zoom call line by line. 

@joachimweyl  ^^

PR #159, opened 2025-03-05
Implemented processor to fetch Coldfront allocation data

### LINKED ISSUE #153: Implement processor to fetch project information from Coldfront
Instead of having coldfront information fetched from our billing scripts earlier in the pipeline, we believe it is more fitting to have this information fetched from the `invoicing` repo instead, contained within a processor class.

This issue is also to allow us to update the way we are currently querying the Coldfront allocations API. We are querying ALL allocations, and this method will not scale well as the NERC grows.

QuanMPhm:
Closes #153. Authentication to Coldfront must be done through Keycloak by providing the client id and secret. These can be set through env vars (`KEYCLOAK_CLIENT_ID` and `KEYCLOAK_CLIENT_SECRET`). A Keycloak and Coldfront url can optionally be provided. They default to the ones hosted by the MGHPCC.

The processor is responsible for adding the PI email and institute code to invoices.
If a Coldfront allocation for an invoice project could not be found, and this project also happens to be
billable, a warning will be logged.

**The query to the Coldfront allocation API will only ask for allocations that appeared on the invoice**

Test cases have been added.

### DIFF
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -3,6 +3,7 @@ import sys
 import datetime
 import logging
 from decimal import Decimal
+import os
 
 import pandas
 import pyarrow
@@ -20,6 +21,7 @@ from process_report.invoices import (
     prepay_credits_snapshot,
 )
 from process_report.processors import (
+    coldfront_fetch_processor,
     validate_pi_alias_processor,
     add_institution_processor,
     lenovo_processor,
@@ -61,6 +63,8 @@ PI_S3_FILEPATH = "PIs/PI.csv"
 ALIAS_S3_FILEPATH = "PIs/alias.csv"
 PREPAY_DEBITS_S3_FILEPATH = "Prepay/prepay_debits.csv"
 
+REQUIRED_ENV_VARS = ("KEYCLOAK_CLIENT_ID", "KEYCLOAK_CLIENT_SECRET")
+
 
 logger = logging.getLogger(__name__)
 logging.basicConfig(level=logging.INFO)
@@ -94,6 +98,12 @@ def load_prepay_csv(prepay_credits_path, prepay_projects_path, prepay_contacts_p
 
 def get_iso8601_time():
     return datetime.datetime.now().strftime("%Y%m%dT%H%M%SZ")
+
+
+def validate_required_env_vars():
+    for required_env_var in REQUIRED_ENV_VARS:
+        if required_env_var not in os.environ:
+            sys.exit(f"Required environment variable {required_env_var} is not set")
 
 
 def main():
@@ -220,6 +230,8 @@ def main():
     )
     args = parser.parse_args()
 
+    validate_required_env_vars()
+
     invoice_month = args.invoice_month
 
     csv_files = args.csv_files or fetch_s3_invoices(invoice_month)
@@ -258,8 +270,13 @@ def main():
 
     ### Preliminary processing
 
+    coldfront_fetch_proc = coldfront_fetch_processor.ColdfrontFetchProcessor(
+        "", invoice_month, merged_dataframe, projects
+    )
+    coldfront_fetch_proc.process()
+
     validate_pi_alias_proc = validate_pi_alias_processor.ValidatePIAliasProcessor(
-        "", invoice_month, merged_dataframe, alias_dict
+        "", invoice_month, coldfront_fetch_proc.data, alias_dict
     )
     validate_pi_alias_proc.process()
 
--- a/process_report/processors/coldfront_fetch_processor.py
+++ b/process_report/processors/coldfront_fetch_processor.py
@@ -1,0 +1,106 @@
+import os
+import sys
+import functools
+import logging
+from dataclasses import dataclass
+
+import requests
+
+from process_report.invoices import invoice
+from process_report.processors import processor
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+
+CF_ATTR_ALLOCATED_PROJECT_NAME = "Allocated Project Name"
+
+
+@dataclass
+class ColdfrontFetchProcessor(processor.Processor):
+    nonbillable_projects: list[str]
+
+    @functools.cached_property
+    def coldfront_client(self):
+        keycloak_url = os.environ.get("KEYCLOAK_URL", "https://keycloak.mss.mghpcc.org")
+
+        # Authenticate with Keycloak
+        token_url = f"{keycloak_url}/auth/realms/mss/protocol/openid-connect/token"
+        r = requests.post(
+            token_url,
+            data={"grant_type": "client_credentials"},
+            auth=requests.auth.HTTPBasicAuth(
+                os.environ["KEYCLOAK_CLIENT_ID"],
+                os.environ["KEYCLOAK_CLIENT_SECRET"],
+            ),
+        )
+        try:
+            r.raise_for_status()
+        except requests.HTTPError:
+            sys.exit(f"Keycloak authentication failed:\n{r.status_code} {r.text}")
+
+        client_token = r.json()["access_token"]
+
+        session = requests.session()
+        headers = {
+            "Authorization": f"Bearer {client_token}",
+            "Content-Type": "application/json",
+        }
+        session.headers.update(headers)
+        return session
+
+    def _get_project_list(self):
+        return self.data[invoice.PROJECT_FIELD].unique()
+
+    def _fetch_coldfront_allocation_api(self, allocation_list):
+        coldfront_api_url = os.environ.get(
+            "COLDFRONT_URL", "https://coldfront.mss.mghpcc.org/api/allocations"
+        )
+        api_query_str = "&".join(
+            [
+                f"attr_{CF_ATTR_ALLOCATED_PROJECT_NAME}={project}"
+                for project in allocation_list
+            ]
+        )
+        r = self.coldfront_client.get(f"{coldfront_api_url}?{api_query_str}")
+
+        return r.json()
+
+    def _get_allocation_data(self, coldfront_api_data):
+        allocation_data = {}
+        for project, project_dict in coldfront_api_data.items():
+            allocation_data[project] = {
+                invoice.PI_FIELD: project_dict["project"]["pi"],
+                invoice.INSTITUTION_ID_FIELD: project_dict["attributes"][
+                    "Institution-Specific Code"
+                ],
+            }
+        return allocation_data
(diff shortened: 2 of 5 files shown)


process_report/processors/coldfront_fetch_processor.py
```diff
@@ -0,0 +1,104 @@
...
+        coldfront_api_url = os.environ.get(
+            "COLDFRONT_URL", "https://coldfront.mss.mghpcc.org/api/allocations"
+        )
+        api_query_str = "&".join(
+            [
+                f"attr_{CF_ATTR_ALLOCATED_PROJECT_NAME}={project}"
+                for project in allocation_list
+            ]
+        )
+        r = self.coldfront_client.get(f"{coldfront_api_url}?{api_query_str}")
```
naved001, line 65: so it looks like instead of asking for all allocations, we are going to pass in a giant list of all the allocations in our report. Is this really that much more scalable?

knikolla: @naved001 That's a valid concern. I believe it is pretty scalable up to a certain point.

Each allocation will require 28 characters for `attr_Allocated Project Name=` + 64 characters for max allocation name length <= 100 characters total per allocation query.

Assuming 100 allocations active * 100 per allocation query = 10 000.

At 1byte per KB that's 10kb, which is already much less in size than the complete list of all allocations.

My only concern would be the whether nginx ingress controller is configured or supports that size of URIs out of the box.

@QuanMPhm can you do some edge testing in your openshift cluster with some really high number of allocations? say 100 and then 1000?

process_report/processors/coldfront_fetch_processor.py
```diff
@@ -0,0 +1,104 @@
...
+                data={"grant_type": "client_credentials"},
+                auth=requests.auth.HTTPBasicAuth(
+                    os.environ["KEYCLOAK_CLIENT_ID"],
+                    os.environ["KEYCLOAK_CLIENT_SECRET"],
+                ),
+            )
+        except KeyError:
+            logger.error(
+                "Keycloak authentication failed. Please set the environment variables KEYCLOAK_CLIENT_ID and KEYCLOAK_CLIENT_ID"
+            )
```
larsks, line 40: You log an error here, but then just keep moving forward. Is that intentional, or should you abort the operation here?

process_report/processors/coldfront_fetch_processor.py
```diff
@@ -0,0 +1,104 @@
...
+        token_url = f"{keycloak_url}/auth/realms/mss/protocol/openid-connect/token"
+        try:
+            r = requests.post(
+                token_url,
+                data={"grant_type": "client_credentials"},
+                auth=requests.auth.HTTPBasicAuth(
+                    os.environ["KEYCLOAK_CLIENT_ID"],
+                    os.environ["KEYCLOAK_CLIENT_SECRET"],
+                ),
+            )
```
larsks, line 36: You never check for HTTP errors. One option is to call `r.raise_for_status()` inside the `try/except` block, and then add an `except` stanza that will translate http errors into a more user friendly error message.

process_report/processors/coldfront_fetch_processor.py
larsks: If these environment variables are *required*, should we be checking for them at an earlier point in program execution?

process_report/processors/coldfront_fetch_processor.py
QuanMPhm: @knikolla I wrote a script that queried a coldfront deployment running in my local Openshift cluster. When quyerying 100 allocations (i.e `/allocations?allo=1&allo=2&...`), the query took 0.3s. At 1000 allocations, the query took 2.3s. Any more allocations beyond that, and we hit [Django's default query field limit](https://docs.djangoproject.com/en/5.1/ref/settings/#data-upload-max-number-fields).

process_report/processors/coldfront_fetch_processor.py
naved001: @QuanMPhm ^^ Please check for errors and log a useful error message.

Somebody accidentally changed my secret (which sets the KEYCLOAK_CLIENT* env variables) and I was wondering why my openshift report was failing. If it logged the http message (in my case 401 for unauthorized) and the client_id I would have immediately spotted the problem instead of debugging the code in the container because it worked fine on my local machine.

process_report/processors/coldfront_fetch_processor.py
QuanMPhm: I will add a function in `process_report.process_report.py` that will check for all required env vars. I will have this function called at the start of `main()` before the s3 input files are fetched.

process_report/process_report.py
```diff
@@ -95,6 +99,12 @@ def get_iso8601_time():
     return datetime.datetime.now().strftime("%Y%m%dT%H%M%SZ")
 
 
+def validate_required_env_vars():
```
larsks, line 102: This function appears to be untested.

larsks (approved):

process_report/process_report.py
QuanMPhm: I added a unit test in `test_util.py`, though given your approval, should I understand that this suggestion was optional?

Also, if you don't mind me rambling on, when does a piece of code become complex enough to warrant a test? I thought this function was too simple to warrant a test. Also, within this repo and a few other repos I worked on, not every function has a test case.

larsks: In an interpreted, dynamically typed language like Python, if your code doesn't execute during your tests then it could be hiding errors that will crop up at runtime. Consider what would happen if `validate_required_env_vars` looked like this:

```
def validate_required_env_vars():
    for required_env_var in REQUIRED_ENV_VAR:
        if required_env_var not in os.environ:
            sys.exit(f"Required environment variable {required_env_var} is not set")
```

There's a typo there -- `REQUIRED_ENV_VARS` is mis-spelled. As it stands, all your tests pass despite this error. That allows it to sit there undetected until somebody actually runs the code.

The goal of unit tests is to prevent this situation.

You don't necessarily need a test for every individual function, but you do need to ensure that as much code is possible is run as part of your tests. This is where tools like `coverage` come in very handy; if we run the tests using `pytest` and the `coverage` plugin...

```
pytest --cov=process_report --cov-report=html
```

...then we get a nice HTML report that shows exactly what lines of code were covered by tests and what lines were not.

larsks (changes requested):
There is a problem with one of the tests. Occasionally, `process_report/tests/unit/processors/test_coldfront_fetch_processor.py::TestColdfrontFetchProcessor::test_coldfront_project_not_found` fails with the following error:

> FAILED process_report/tests/unit/processors/test_coldfront_fetch_processor.py::TestColdfrontFetchProcessor::test_coldfront_project_not_found - AssertionError: "Projects {'P5', 'P4'} not found in Coldfront and are billable! Please check the project names" not found in "WARNING:process_report.processors.coldfront_fetch_processor:Projects {'P4', 'P5'} not found in Coldfront and are billable! Please check the project names"

~It's too late at night for me to figure out what's going on here, but~ when I ran this test 100 times just now it failed 3 times:

```
$ ok=0; failed=0; for _ in {1..100}; do if uv run pytest process_report/tests/unit/processors/test_coldfront_fetch_processor.py; then let ok++; else let failed++; fi ; done; echo ok $ok failed $failed
ok 97 failed 3
```

On a second run, it worked 99 times and failed 1 time.

We should figure out what's going on here. This is either a buggy test, which will lead to erroneous CI failures, or it's an actual bug in the code.

---

Update: it's an ordering error. When the test fails, `log.output[0]` contains:

```
(Pdb) log.output[0]
"WARNING:process_report.processors.coldfront_fetch_processor:Projects {'P4', 'P5'} not found in Coldfront and are billable! Please check the project names"
```

But we're checking it against:

```
(Pdb) f"Projects {answer_project_set} not found in Coldfront and are billable! Please check the project names"
"Projects {'P5', 'P4'} not found in Coldfront and are billable! Please check the project names"
```

You need to ensure that the list of projects in the error message is always presented in the same order.

process_report/processors/coldfront_fetch_processor.py
```diff
@@ -0,0 +1,106 @@
...
+CF_ATTR_ALLOCATED_PROJECT_NAME = "Allocated Project Name"
+
+
+@dataclass
+class ColdfrontFetchProcessor(processor.Processor):
+    nonbillable_projects: list[str]
+
+    @functools.cached_property
+    def coldfront_client(self):
+        keycloak_url = os.environ.get("KEYCLOAK_URL", "https://keycloak.mss.mghpcc.org")
```
larsks, line 25: I'm not suggesting that you change this now, but in general a good design principle is get configuration from the environment *outside* of the class (e.g., in your `main` method), and then pass that information into the class as class parameters. This makes things easier to test (you can just instantiate classes with your choice of values), and leads to better reuse (what if want to get values from a configuration file rather than the environment?).

larsks (approved):

QuanMPhm:
@knikolla @naved001 If it's fine, I would like your input before merging

process_report/tests/unit/processors/test_coldfront_fetch_processor.py
```diff
@@ -0,0 +1,92 @@
...
+        return mock_data
+
+    @mock.patch(
+        "process_report.processors.coldfront_fetch_processor.ColdfrontFetchProcessor._fetch_coldfront_allocation_api",
+    )
+    def test_coldfront_fetch(self, mock_get_allocation_data):
+        mock_get_allocation_data.return_value = self._get_mock_allocation_data(
+            ["P1", "P2", "P3", "P4"],
+            ["PI1", "PI1", "", "PI12"],
+            ["IC1", "", "", "IC2"],
```
naved001, line 46: Just so I understand: you are associating `PI1` with Project `P2` but then it doesn't have an Institution Code. Is the institution code associated with the Project or the PI? Probably doesn't matter for the test, but I would like it to match the real world data.

naved001 (approved):

process_report/tests/unit/processors/test_coldfront_fetch_processor.py
QuanMPhm: From the invoicing requirements file, the institute code is unique to each project, so I guess this test data could be real world data.

QuanMPhm:
This is dependant on both https://github.com/CCI-MOC/openshift-usage-scripts/issues/111 and https://github.com/CCI-MOC/openstack-billing-from-db/issues/91

PR #161, opened 2025-03-11
Used Pydantic to validate `institute_list.yaml`

### LINKED ISSUE #158: Implement validation for `institute_list.yaml`
#157 was caused by a typo in the key of a YAML object. This is something we can implement validation for to prevent it happening again.

- [ ] Implement validation
- [ ] Have it run as part of CI

QuanMPhm:
Closes #158. More details in the commit messages. Thank you to @larsks, since most of this work was derived from his [code](https://github.com/CCI-MOC/nerc-rates/blob/main/src/nerc_rates/cmd/validate_rates_file.py) in `nerc-rates`.

### DIFF
--- a/.github/workflows/validate-institute-list.yaml
+++ b/.github/workflows/validate-institute-list.yaml
@@ -22,4 +22,4 @@ jobs:
 
       - name: Validate institute list
         run: |
-          python -m process_report.instiute_list_validate --github
+          python -m process_report.institute_list_validate --github
--- a/.github/workflows/validate-institute-list.yaml
+++ b/.github/workflows/validate-institute-list.yaml
@@ -1,0 +1,25 @@
+name: Validate institute list
+on:
+  push:
+  pull_request:
+
+jobs:
+  run-unit-tests:
+    name: Validate institute list
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Set up python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+
+      - name: Install dependencies
+        run: |
+          pip install -r requirements.txt
+
+      - name: Validate institute list
+        run: |
+          python -m process_report.instiute_list_validate --github
--- a/process_report/institute_list_models.py
+++ b/process_report/institute_list_models.py
@@ -1,0 +1,61 @@
+from typing import Annotated
+import datetime
+
+import pydantic
+import validators
+
+
+def parse_date(v: str) -> str:
+    try:
+        datetime.datetime.strptime(v, "%Y-%m")
+        return v
+    except ValueError:
+        raise ValueError(f"Invalid date string {v}. Must be in format YYYY-MM")
+
+
+def validate_domain(v: str) -> str:
+    if not validators.domain(
+        v, consider_tld=True
+    ):  # Ensures only TLDs allowed by IANA are valid
+        raise ValueError(f"Invalid domain {v}")
+    return v
+
+
+DateField = Annotated[str, pydantic.BeforeValidator(parse_date)]
+DomainField = Annotated[str, pydantic.AfterValidator(validate_domain)]
+
+
+class InstituteInfo(pydantic.BaseModel):
+    display_name: str
+    domains: list[DomainField]
+    mghpcc_partnership_start_date: DateField | None = None
+    include_in_nerc_total_invoice: bool = False
+
+    model_config = pydantic.ConfigDict(extra="forbid")
+
+
+class InstituteList(pydantic.RootModel):
+    root: list[InstituteInfo]
+
+    @pydantic.model_validator(mode="after")
+    def validate_no_display_name_duplicates(self):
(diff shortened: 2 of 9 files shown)


.github/workflows/validate-institute-list.yaml
```diff
@@ -0,0 +1,25 @@
...
+        with:
+          python-version: '3.10'
+
+      - name: Install dependencies
+        run: |
+          pip install -r requirements.txt
+
+      - name: Validate institute list
+        run: |
+          python -m process_report.instiute_list_validate --github
```
larsks, line 25: This looks like typo (`instiute_list_validate` should be `institute_list_validate`).

process_report/institute_list_models.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+DateField = Annotated[str, pydantic.BeforeValidator(parse_date)]
+DomainField = Annotated[str, pydantic.AfterValidator(validate_domain)]
+
+
+class InstituteInfo(pydantic.BaseModel):
+    display_name: str
+    domains: list[DomainField]
+    mghpcc_partnership_start_date: DateField = None
```
larsks, line 31: This is an invalid specification, because `None` is not a `DateField`. You need:

```suggestion
    mghpcc_partnership_start_date: DateField|None = None
```

process_report/institute_list_models.py
```diff
@@ -0,0 +1,61 @@
...
+    include_in_nerc_total_invoice: bool = None
```
larsks, line 32: Similar to the above, although in this case you could instead write:

```suggestion
    include_in_nerc_total_invoice: bool = False
```

larsks (changes requested):

.github/workflows/validate-institute-list.yaml
QuanMPhm: Indeed it is :P

process_report/institute_list_models.py
```diff
@@ -0,0 +1,61 @@
+from typing import Annotated
+import datetime
+
+import pydantic
+import validators
```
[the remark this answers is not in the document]
QuanMPhm: I used to [here](https://github.com/CCI-MOC/invoicing/blob/3ba504bc17cd6d1cc22a488054b3ad3f3ae3e2cc/process_report/institute_list_models.py#L17) to validate the domain urls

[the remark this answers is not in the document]
larsks: I think you mis-understoof my comment. **There is no `validators` module**:

```
$ python -m process_report.institute_list_validate
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/lars/redhat/moc/invoicing/process_report/institute_list_validate.py", line 6, in <module>
    from process_report.institute_list_models import InstituteList
  File "/home/lars/redhat/moc/invoicing/process_report/institute_list_models.py", line 5, in <module>
    import validators
ModuleNotFoundError: No module named 'validators'
```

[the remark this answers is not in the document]
QuanMPhm: Ooooh. Haha I see. I noticed that too when seeing why the workflow failed. I just amended the PR to fix that

process_report/institute_list_models.py
larsks: Pydantic should have yelled at you if you attempted to instantiate these objects using default values. If it didn't, maybe we need to add an additional test.

QuanMPhm: I just tested `Pydantic` too, and it seems if the field is not present in the YAML file, `Pydantic` is fine with the field defaulting to `None`. I'm fine writing a unit test for the validator, or a validator for the validator.

larsks: I guess it's just type checkers (like mypy or basedpyright) that complain:

```
  /home/lars/redhat/moc/invoicing/process_report/institute_list_models.py:31:48 - error: Type "None" is not assignable to declared type "DateField"
    "None" is not assignable to "str" (reportAssignmentType)
```

process_report/institute_list_validate.py
```diff
@@ -0,0 +1,47 @@
...
+    try:
+        with open(args.institute_list) as f:
+            data = yaml.safe_load(f)
+            institute_list = InstituteList.model_validate(data)
+            print(
+                f"Validation of {len(institute_list.root)} institution entries successful"
+            )
+    except pydantic.ValidationError as err:
+        if args.github:
+            pydantic_to_github(err, args.rates_file)
```
larsks, line 42: There is no argument named `rates_file`.

QuanMPhm: I've added a simple test case for the validator

larsks (approved):

joachimweyl:
@QuanMPhm is this ready to merge?

QuanMPhm:
@joachimweyl I would like to at least have a review from @knikolla or @naved001

naved001 (approved):

PR #163, opened 2025-03-12
Don't create test directories in current working directory

larsks:
The `test_export_pi` test would leave an empty `test_dir` in the current
directory. We fix that by wrapping the relevant code in a
`tempfile.TemporaryDirectory` context manager, which creates a temporary
directory and cleans it up when the context completes.


QuanMPhm (approved):

QuanMPhm:
Thank you @larsks. I am surprised I did not used `tempfile` when writing that unit test!

PR #170, opened 2025-04-07
Fix temporary file unit test errors

KelvinLinBU:
Fixing of temp file errors within unit tests. Ensure proper closing and access to temp files for any OS that lock files by default (this prevents any process from opening it for reading while the file is open). Should allow for unit tests to run on more machines while maintaining original function.

### DIFF
--- a/process_report/tests/unit/test_institute_list_validate.py
+++ b/process_report/tests/unit/test_institute_list_validate.py
@@ -20,7 +20,7 @@ class TestInstituteListValidate(TestCase):
             },
         ]
 
-        with tempfile.NamedTemporaryFile(mode="w") as f:
+        with tempfile.NamedTemporaryFile(mode="w", delete=False) as f:
             yaml.dump(test_institute_list, f)
             f.flush()
             main(["--github", f.name])
@@ -35,7 +35,7 @@ class TestInstituteListValidate(TestCase):
             }
         ]
 
-        with tempfile.NamedTemporaryFile(mode="w") as f:
+        with tempfile.NamedTemporaryFile(mode="w", delete=False) as f:
             yaml.dump(test_institute_list, f)
             f.flush()
             with self.assertRaises(SystemExit):
--- a/process_report/tests/unit/processors/test_new_pi_credit_processor.py
+++ b/process_report/tests/unit/processors/test_new_pi_credit_processor.py
@@ -110,6 +110,7 @@ class TestNewPICreditProcessor(TestCase):
         self.test_old_pi_file = tempfile.NamedTemporaryFile(
             delete=False, mode="w+", suffix=".csv"
         )
+        self.test_old_pi_file.close()
 
     def tearDown(self) -> None:
         os.remove(self.test_old_pi_file.name)
--- a/process_report/tests/unit/processors/test_prepayment_processor.py
+++ b/process_report/tests/unit/processors/test_prepayment_processor.py
@@ -89,6 +89,7 @@ class TestPrepaymentProcessor(TestCase):
         self.test_prepay_debits_file = tempfile.NamedTemporaryFile(
             delete=False, mode="w+", suffix=".csv"
         )
+        self.test_prepay_debits_file.close()
 
     def tearDown(self) -> None:
         os.remove(self.test_prepay_debits_file.name)


process_report/tests/unit/test_institute_list_validate.py
```diff
@@ -20,7 +20,7 @@ def test_valid_institute_list(self):
             },
         ]
 
-        with tempfile.NamedTemporaryFile(mode="w") as f:
+        with tempfile.NamedTemporaryFile(mode="w", delete=False) as f:
```
QuanMPhm, line 23: Why are we not deleting these files?

process_report/tests/unit/test_institute_list_validate.py
```diff
@@ -35,7 +35,7 @@ def test_invalid_institute_list(self):
             }
         ]
 
-        with tempfile.NamedTemporaryFile(mode="w") as f:
+        with tempfile.NamedTemporaryFile(mode="w", delete=False) as f:
```
QuanMPhm, line 38: ^

QuanMPhm (changes requested):
A small question

process_report/tests/unit/test_institute_list_validate.py
KelvinLinBU: Ah, the way I understand this is that Windows locks files while they're open in Python's NamedTemporaryFile(), so if you try to reuse the file path too early, you get a permission error. By using delete=False and waiting until the with block ends (so the file closes), it works because the file still exists and the lock is released.

QuanMPhm: In any case, the temp files should be deleted. The `tempfile` [documentation](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) suggests the following:

> Opening the temporary file again by its name while it is still open works as follows:
On POSIX the file can always be opened again.
On Windows, make sure that at least one of the following conditions are fulfilled:
> - delete is false
> - additional open shares delete access (e.g. by calling [os.open()](https://docs.python.org/3/library/os.html#os.open) with the flag O_TEMPORARY)
> - delete is true but delete_on_close is false. Note, that in this case the additional opens that do not share delete access (e.g. created via builtin [open()](https://docs.python.org/3/library/functions.html#open)) must be closed before exiting the context manager, else the [os.unlink()](https://docs.python.org/3/library/os.html#os.unlink) call on context manager exit will fail with a [PermissionError](https://docs.python.org/3/library/exceptions.html#PermissionError).

Try their suggestion and check if the temp files are deleted after the tests

QuanMPhm:
@KelvinLinBU After asking my senior, I would recommend that you continue development on a Linux environment rather than Windows. As far as I am aware, most of our codebase is written and tested on Linux environments. For now, it is not practical for us to add support for Windows, and developing in Windows will only slow you down in the time being.

I recommend you look into [Windows Subsystem for Linux](https://learn.microsoft.com/en-us/windows/wsl/about) if you don't have easy access to a Linux machine.

That being said, I'll close this PR.

larsks:
I think the code involving temporary files here is in need of fixing, but it's certainly not a high priority. In almost all cases, I would prefer that we don't open the file in the `setUp` method, and instead scope the temporary files to individual functions.

I think the best way of doing this would be to abandon the current test structure and rewrite things using [pytest](https://docs.pytest.org/en/stable/); pytest's [fixtures](https://docs.pytest.org/en/6.2.x/fixture.html) support makes it very easy to provide function-scoped, automatically cleaned up test resources...but that's a major change and not one I would suggest pursuing at this time.

Looking at the existing code, it seems like in most cases we could use a temporary directory instead:

```
import tempfile
import shutil
from pathlib import Path

class TestSomething(TestCase):
    def setUp(self):
        self.tempdir = Path(tempfile.TemporaryDirectory(delete=False).name)

    def test_something(self):
        test_old_pi_file = self.tempdir / "old_pi_file"

        test_old_pi_df = pandas.DataFrame(
            {
                "PI": ["PI"],
                "First Invoice Month": ["2024-01"],
                "Initial Credits": [1000],
            }
        )
        test_old_pi_df.to_csv(test_old_pi_file, index=False)

        ...

    def tearDown(self):
        shutils.rmtree(self.tempdir)
```

That would I think resolve the concerns this PR was meant to address (we would never be trying to open a file that was already open).

You could make the temp directory setup/teardown part of a base class that gets inherited by all your test cases.

KelvinLinBU:
@QuanMPhm, what are your thoughts on @larsks suggestion?

QuanMPhm:
@KelvinLinBU Feel free to make another PR that implements a base class with setup/teardown for temp dirs as @larsks suggested.

PR #173, opened 2025-04-10
Add depedantbot configuration to keep actions up-to-date

larsks:
This will automate the process of keeping actions in our github workflows
up-to-date. It runs periodically, and when a new action version is
available, dependabot will generate a pull request with the necessary
changes.


QuanMPhm:
@larsks I guess people will make things for everything

QuanMPhm (approved):

PR #177, opened 2025-04-14
Fix unit test to avoid opening of already opened files

KelvinLinBU:
Added class BaseTestCaseWithTempDir with custom setUp and tearDown functions compared to regular TestCase. Adjusted tests per suggestion in https://github.com/CCI-MOC/invoicing/pull/170. Removal of replaced functions and unneeded imports after change.


KelvinLinBU:
@QuanMPhm

process_report/tests/unit/processors/test_new_pi_credit_processor.py
```diff
@@ -202,7 +193,7 @@ def test_one_new_pi(self):
         self._assert_result_invoice_and_old_pi_file(
             invoice_month,
             test_invoice,
-            self.test_old_pi_file.name,
+            test_old_pi_file,
```
QuanMPhm, line 196: While passing in a `Path` object does technically work `[pandas.read_csv](https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html)` does accept Path-like objects, our invoice and processor classes [expect the filepaths to be a str](https://github.com/CCI-MOC/invoicing/blob/211076bf5e6405917255c5b20f3ce7d3c22db133/process_report/processors/new_pi_credit_processor.py#L28). Please update your PR to make sure the filepath str is passed to the processors instead.

QuanMPhm (changes requested):
You should combine your two commits into one because they are very related to one another. In fact, the first commit by itself functionally does nothing, and only becomes used after the second commit. 

You should consider each commit as a change that "stands on its own". Each commit should do one thing, and do it clearly. Having one change (in this case, refactoring unit tests to avoid opening already opened files) split into more than one commit makes its a bit more confusing how each each commit related to one another (suppose someone in the future review this repo's commit history, its not immediately obvious that the second commit depends on the first commit, or that both of them are needed to implement a single feature)

QuanMPhm (approved):

PR #189, opened 2025-05-06
Update institute_list.yaml

joachimweyl:


joachimweyl:
@QuanMPhm any idea why the first 4 tests failed?

QuanMPhm:
@joachimweyl Bentley university [already has an entry](https://github.com/CCI-MOC/invoicing/blob/ccca1c6a26782329489f16ddaa2fb3ae19b5882d/process_report/institute_list.yaml#L10-L12) in the institute list. I actually feel pretty giddy that my test has actually caught an honest error.

neclinton:
Yeah, it works.  We hadn't charged Bentley yet, so that will now start.  that's why I initiated this.

PR #195, opened 2025-05-12
Added CLI option to pass in Coldfront API data

### LINKED ISSUE #194: Allow passing Coldfront data through CLI or bypass Coldfront processor
Currently, the Coldfront processor will throw an error if it can't communicate with a Coldfront API, or if the Coldfront data doesn't contain enough PI info. This makes it impossible to test the invoicing system locally

We should add a CLI argument that allows passing in custom Coldfront data, or add a toggle to bypass the processor.

QuanMPhm:
Closes #194. The filepath to a JSON file can be now passed through the CLI. This will allow injection of custom Coldfront data for ease of testing purposes.

### DIFF
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -63,8 +63,6 @@ PI_S3_FILEPATH = "PIs/PI.csv"
 ALIAS_S3_FILEPATH = "PIs/alias.csv"
 PREPAY_DEBITS_S3_FILEPATH = "Prepay/prepay_debits.csv"
 
-REQUIRED_ENV_VARS = ("KEYCLOAK_CLIENT_ID", "KEYCLOAK_CLIENT_SECRET")
-
 
 logger = logging.getLogger(__name__)
 logging.basicConfig(level=logging.INFO)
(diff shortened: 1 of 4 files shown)


process_report/process_report.py
```diff
@@ -164,6 +164,11 @@ def main():
         default="prepaid_contacts.csv",
         help="CSV listing all prepay group contact information. Defaults to 'prepaid_contacts.csv'",
     )
+    parser.add_argument(
+        "--coldfront-data-file",
+        required=False,
+        help="JSON file containing coldfront allocation data, must follow the format of the coldfront-plugin-api API",
+    )
```
naved001, line 171: So, is this purely for test purposes? because if I were to call `main()` but don't set these:

```
REQUIRED_ENV_VARS = ("KEYCLOAK_CLIENT_ID", "KEYCLOAK_CLIENT_SECRET")
```

Then it will complain even though there's no reason to set those if I am providing a file. Obviously as a workaround I could just set these to whatever.

process_report/processors/coldfront_fetch_processor.py
```diff
@@ -62,6 +64,13 @@ def _fetch_coldfront_allocation_api(self):
 
         return r.json()
 
+    def _get_coldfront_api_data(self):
+        if self.coldfront_data_filepath:
```
naved001, line 68: maybe log an info message that we are reading the data from a file at the given path. This is just in case someone provides a file path AND the environment variables for coldfront_url/keycloak_stuff.

naved001 (commented):
So, if

process_report/process_report.py
QuanMPhm: Ah good point. I'll make the env vars optional if the `--coldfront-data-file` option is set

naved001 (approved):

Issue #167, opened 2025-03-25
Implement invoice for ocp-test


QuanMPhm:
Related to https://github.com/CCI-MOC/openshift-usage-scripts/issues/108#issuecomment-2751540998, we want to generate an invoice for the test cluster. This cluster has no special invoicing requirements, so it can simply be implemented as a simple `Invoice` subclass.

joachimweyl:
@QuanMPhm I thought this was already done.

QuanMPhm:
@joachimweyl @knikolla After thinking a bit, I now realize that we need to add every project that could appear in the `ocp-test` invoice into the nonbillable list. This is needed because:
1. Otherwise, the Coldfront processor would raise an [error](https://github.com/CCI-MOC/invoicing/blob/7d78602b7e7a99adee4f848211ad64537b111e35/process_report/processors/coldfront_fetch_processor.py#L98)
2. PIs with projects in `ocp-test` may lose some of their New-PI credit
3. Probably the biggest problem, prepayments would be applied to these projects.

Let me know what you think before I write up the issue in the nonbillable repo

joachimweyl:
makes sense. list created.

QuanMPhm:
From today's 2:30 meeting discussion, it was agreed that it's fine to hardcode `ocp-test` as a non-billable cluster. With regards to the one billable project (`grits-grots`) in `ocp-test`, @joachimweyl said that he can handle that project manually

QuanMPhm:
Blocked by #196

PR #196, opened 2025-05-12
Added nonbillable cluster list to billable processor

### LINKED ISSUE #165: Add logic to handle a `Cluster Name` column in invoicing
During discussions, we found the need to specify which clusters a certain project invoice originated from, whether that be from an Openstack, Openshift, or Bare Metal server. This column would be used internally to provide cluster-specific processing  (i.e a credit for a specific cluster)

The list of cluster names we agreed on was:
- Openstack: `stack`
- Openshift production: `ocp-prod`
- Virtualization cluster: `virt`
- OpenShift Test: `ocp-test`
- OpenShift Academic: `academic`
- Bare Metal: `bm`

@joachimweyl @naved001 Please confirm these are the agreed cluster names. I'll write the related issues on the other billing repos

QuanMPhm:
Closes #165. More details in the commit message. This is blocking #167 and #152.

I have a feeling this solution for the `ocp-test` debacle is a bit clunky. Let me know what you guys think.

### DIFF
--- a/process_report/tests/unit/processors/test_coldfront_fetch_processor.py
+++ b/process_report/tests/unit/processors/test_coldfront_fetch_processor.py
@@ -11,6 +11,7 @@ class TestColdfrontFetchProcessor(TestCase):
         allocation_project_name=None,
         pi=None,
         institute_code=None,
+        cluster_name=None,
     ):
         if not pi:
             pi = [""] * len(allocation_project_id)
@@ -19,7 +20,10 @@ class TestColdfrontFetchProcessor(TestCase):
             institute_code = [""] * len(allocation_project_id)
 
         if not allocation_project_name:
-            allocation_project_name = allocation_project_id.copy()
+            allocation_project_name = [""] * len(allocation_project_id)
+
+        if not cluster_name:
+            cluster_name = [""] * len(allocation_project_id)
 
         return pandas.DataFrame(
             {
@@ -27,6 +31,7 @@ class TestColdfrontFetchProcessor(TestCase):
                 "Project - Allocation": allocation_project_name,
                 "Project - Allocation ID": allocation_project_id,
                 "Institution - Specific Code": institute_code,
+                "Cluster Name": cluster_name,
             }
         )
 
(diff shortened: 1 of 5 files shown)


process_report/tests/unit/processors/test_coldfront_fetch_processor.py
```diff
@@ -95,3 +100,31 @@ def test_coldfront_project_not_found(self, mock_get_allocation_data):
             str(cm.exception),
             f"Projects {answer_project_set} not found in Coldfront and are billable! Please check the project names",
         )
+
+    @mock.patch(
+        "process_report.processors.coldfront_fetch_processor.ColdfrontFetchProcessor._fetch_coldfront_allocation_api",
+    )
+    def test_nonbillable_clusters(self, mock_get_allocation_data):
+        """What happens when an invoice project is not found in Coldfront."""
```
naved001, line 108: So how is this testing nonbillable clusters? it just seems to be testing projects that don't have information in coldfront. Would it have mattered if in your test invoice the cluster name wasn't `ocp-test`?

QuanMPhm: The Coldfront processor throws an error if an invoice project is 1. Not found in Coldfront, and 2. Is billable.
This test checks that even though Coldfront has no info on Projects `P3` and `P4`, it did not throw an error because these projects belong to the nonbillable `ocp-test` cluster.

If `P3` and `P4` did not belong to `ocp-test`, an error would be raised

naved001: Thanks for the clarification. Can you update the docstring to be more explicit then? Maybe something like:

"""
No errors are raised when an invoice project belonging to a non billable cluster (ocp-test) is not found in Coldfront
"""

> If P3 and P4 did not belong to ocp-test, an error would be raised

Ah, and we are testing that in `test_coldfront_project_not_found`.

QuanMPhm: The docstring has been amended

naved001 (approved):

PR #197, opened 2025-05-20
Removal of unused functions and variables

KelvinLinBU:
Closes https://github.com/CCI-MOC/invoicing/issues/192. Cleanup of some utility functions and unused variables.

### DIFF
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -1,6 +1,5 @@
 import argparse
 import sys
-import datetime
 import logging
 from decimal import Decimal
 import os
@@ -11,6 +10,7 @@ from nerc_rates import load_from_url
 
 from process_report import util
 from process_report.invoices import (
+    invoice,
     lenovo_invoice,
     nonbillable_invoice,
     billable_invoice,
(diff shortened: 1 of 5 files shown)


QuanMPhm (changes requested):
Looks good. Just a few other spots I think you could clean up

process_report/process_report.py
```diff
@@ -10,6 +9,7 @@
 from nerc_rates import load_from_url
 
 from process_report import util
+from process_report.util import get_iso8601_time
```
QuanMPhm, line 12: This file already imports `util`. You can use `util.get_iso8601_time()` instead

process_report/process_report.py
QuanMPhm, line 1: Can you make this file use the [constants](https://github.com/CCI-MOC/invoicing/blob/54019b2f237e00f8e623aff3ea98a065102920ed/process_report/invoices/invoice.py#L28-L49) defined in `invoice.py`?

QuanMPhm (changes requested):
Some more changes I'd like to have

process_report/process_report.py
```diff
@@ -30,6 +29,11 @@
     bu_subsidy_processor,
     prepayment_processor,
 )
+from process_report.invoices.invoice import (
```
QuanMPhm, line 32: All of our other files import using `from process_report.invoices import invoice`. For uniformity, please use that pattern instead.

QuanMPhm: You also did not include [`INVOICE_DATE_FIELD`](https://github.com/CCI-MOC/invoicing/blob/0a8b8623072b0cb55022b5ed1053fa002eb124f0/process_report/process_report.py#L440). This would have raised an error if the top module was ran.

process_report/process_report.py
```diff
@@ -30,6 +29,11 @@
...
+    RATE_FIELD,
+    COST_FIELD,
+)
+
 
 ### PI file field names
 PI_PI_FIELD = "PI"
```
QuanMPhm, line 39: Can you remove these constants as well?

QuanMPhm (changes requested):

QuanMPhm (changes requested):
Can you remove [this unused cli argument ](https://github.com/CCI-MOC/invoicing/blob/0a8b8623072b0cb55022b5ed1053fa002eb124f0/process_report/process_report.py#L119-L123)as well?

QuanMPhm:
@knikolla @naved001 A relatively simple clean-up PR for `invoicing`. Since @KelvinLinBU is currently busy with other work, I'll wrap up his lingering PRs

knikolla (approved):

PR #200, opened 2025-05-20
Update Ruff to latest version v.0.11.10

KelvinLinBU:
Closes https://github.com/CCI-MOC/invoicing/issues/199. This update will help us not use outdated Ruff software.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.5.0
+    rev: v5.0.0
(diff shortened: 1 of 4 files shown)


QuanMPhm (commented):
I don't have a strong opinion regarding this change. @knikolla @larsks Do you have any thoughts?

larsks (approved):

larsks:
I've updated the PR by running `pre-commit autoupdate`, since if you're going to update one hook you might as well update them all. I've also split it into two commits; one that updates the hook versions, and one that fixes newly reported errors.

QuanMPhm (approved):

PR #201, opened 2025-05-27
Refactor type handling in invoices

### LINKED ISSUE #191: Refactoring how we're handling typing
Currently, our code is quite disorganized with how it is handling typing. A quick search of the term "pyarrow" or "decimal" shows that we're explicitly typing certain fields in miscellaneous places. In a few more places and even in [our test cases](https://github.com/CCI-MOC/invoicing/blob/724dcff9bbac962cff9c8da091a84a31a0719f8e/process_report/tests/unit/processors/test_bu_subsidy_processor.py#L22), we have to typecast to avoid equality checks.

Because of how disorganized things are, this has led to:
1. Debugging typing-related problems leads to even more messiness.
2. Confusion over the typing of fields, which leads to immense frustration when writing test cases

We should re-consider how we're handling typing to reduce future programming burden.

@larsks @naved001 @knikolla Happy to hear your thoughts.

QuanMPhm:
Closes #191. More details in the commit message.

Currently, the test cases are failing because the PI-specific invoice [injects an incompatible value in the invoice dataframe](https://github.com/CCI-MOC/invoicing/blob/54019b2f237e00f8e623aff3ea98a065102920ed/process_report/invoices/pi_specific_invoice.py#L108). After some thinking, I thought the PI invoice could do some refactoring in that spot as well. I'll put this as a draft for now while I work on refactor the PI invoice.

### DIFF
--- a/process_report/invoices/invoice.py
+++ b/process_report/invoices/invoice.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
 import pandas
+import pyarrow
 
 import process_report.util as util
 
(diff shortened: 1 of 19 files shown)


QuanMPhm:
Closing this PR as it is now overshadowed by #285

PR #214, opened 2025-06-27
Consolidate configurations into a `config.py` file

### LINKED ISSUE #213: Consolidate configuration for invoicing
As mentioned by @knikolla and @jimmysway in Slack, the `process_report.py` module contains hundreds of lines of cli arguements, and because each invoice/processor requires different configurations, we are required to explicitly initialize each of them in the top-level module. 

As suggested by @knikolla, it would be best if we could put as much configurations into a `config.py` module, and have everything fetch from it. This would also standardize the instantiation of our invoices/processors:
> - As you mentioned, the main function is several hundreds lines, mostly due to loading configuration.
>  - We can put all configuration options in a config.py. Including defaults as much as possible to require close to zero arguments explicitly be provided to start invoicing.
>  - Processors and Invoices should load all configurations from config.py rather than require __init__ arguments. This allows main() to treat them all in the same way, iterating over a list. In particular Invoice names and output filepaths have no reason to be editable and therefore we can remove all the arguments that pertaining to them.

QuanMPhm:
Closes #213. I'll submit this as a draft for now, since its quite voluminous already, and I want people's opinions before moving forward. 

@knikolla @larsks @naved001 In particular, right now the unit tests are failing because `config.py` tries to fetch from s3 while unauthenticated. I was considering creating a base test class that mocks the s3 functions during `setUp()`. Do you think there's a better way?

### DIFF
--- a/process_report/config.py
+++ b/process_report/config.py
@@ -1,0 +1,137 @@
+import functools
+from datetime import datetime, timedelta
+from decimal import Decimal
+
+import pandas
+from nerc_rates import load_from_url
+
+from process_report import util
+
+
+PI_S3_FILEPATH = "PIs/PI.csv"
+ALIAS_S3_FILEPATH = "PIs/alias.csv"
+PREPAY_DEBITS_S3_FILEPATH = "Prepay/prepay_debits.csv"
+
+
+rates_info = load_from_url()
+
+
+def default_invoice_month():
+    """
+    Start of the invoicing period (YYYY-MM). Defaults to last month if 1st of a month, or this month otherwise.
+    """
+    d = datetime.today() - timedelta(days=1)
+    return d.strftime("%Y-%m")
+
+
+def fetch_s3_invoices(invoice_month):
+    """Fetches usage invoices from S3 given invoice month"""
+    s3_invoice_list = list()
+    invoice_bucket = util.get_invoice_bucket()
+    for obj in invoice_bucket.objects.filter(
+        Prefix=f"Invoices/{invoice_month}/Service Invoices/"
+    ):
+        local_name = obj.key.split("/")[-1]
+        s3_invoice_list.append(local_name)
+        invoice_bucket.download_file(obj.key, local_name)
+
+    return s3_invoice_list
+
+
+INVOICE_MONTH = default_invoice_month()
+UPLOAD_TO_S3 = True
+
+
+### Invoice file names
+NONBILLABLE_INVOICE_NAME = "nonbillable"
+BILLABLE_INVOICE_NAME = "billable"
+BU_INTERNAL_INVOICE_NAME = "BU_Internal"
+NERC_TOTAL_INVOICE_NAME = "NERC"
+LENOVO_INVOICE_NAME = "Lenovo"
+OCP_TEST_INVOICE_NAME = "OCP_TEST"
+PI_SPECIFIC_FOLDER_NAME = "pi_invoices"
+
+### Input file paths
+NONBILLABLE_PIS_FILEPATH = "pi.txt"
+NONBILLABLE_PROJECTS_FILEPATH = "project.txt"
+NONBILLABLE_TIMED_PROECTS_FILEPATH = "timed_project.txt"
+
+PREPAY_PROJECTS_FILEPATH = "prepaid_projects.csv"
+PREPAY_CREDITS_FILEPATH = "prepaid_credits.csv"
+PREPAY_CONTACTS_FILEPATH = "prepaid_contacts.csv"
+
+CSV_INVOICE_FILEPATH_LIST = fetch_s3_invoices(INVOICE_MONTH)
+PREPAY_DEBITS_FILEPATH = util.fetch_s3(PREPAY_DEBITS_S3_FILEPATH)
+OLD_PI_FILEPATH = util.fetch_s3(PI_S3_FILEPATH)
+ALIAS_FILEPATH = util.fetch_s3(ALIAS_S3_FILEPATH)
+COLDFRONT_API_FILEPATH = ""  # Defaults to being fetched in ColdfrontFetchProcessor
+
+
+### Miscellaneous config values
+NEW_PI_CREDIT_AMOUNT = rates_info.get_value_at("New PI Credit", INVOICE_MONTH, Decimal)
+LIMIT_NEW_PI_CREDIT_TO_PARTNERS = rates_info.get_value_at(
+    "Limit New PI Credit to MGHPCC Partners", INVOICE_MONTH, bool
+)
+BU_SUBSIDY_AMOUNT = rates_info.get_value_at("BU Subsidy", INVOICE_MONTH, Decimal)
+
+
+### Getter functions. Don't change these
+@functools.lru_cache
+def get_nonbillable_pis():
+    with open(NONBILLABLE_PIS_FILEPATH) as file:
+        return [line.rstrip() for line in file]
+
+
+@functools.lru_cache
+def get_nonbillable_projects():
+    """Returns list of nonbillable projects for current invoice month"""
+    with open(NONBILLABLE_PROJECTS_FILEPATH) as file:
+        projects = [line.rstrip() for line in file]
+
+    timed_projects_list = get_nonbillable_timed_projects(
+        NONBILLABLE_TIMED_PROECTS_FILEPATH, INVOICE_MONTH
+    )
+    return list(set(projects + timed_projects_list))
+
+
+@functools.lru_cache
+def get_nonbillable_timed_projects(timed_projects_file, invoice_date):
+    """Returns list of projects that should be excluded based on dates"""
+    dataframe = pandas.read_csv(timed_projects_file)
+
+    # convert to pandas timestamp objects
+    dataframe["Start Date"] = pandas.to_datetime(
+        dataframe["Start Date"], format="%Y-%m"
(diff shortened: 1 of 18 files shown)


larsks:
It's not part of this PR, but [this code](https://github.com/QuanMPhm/process_csv_report/blob/d6b61abe0992fd297d3637c189f5db4321023512/process_report/util.py#L19-L33
) in `util.py` is problematic:

```python
@functools.lru_cache
def get_invoice_bucket():
    try:
        s3_resource = boto3.resource(
            service_name="s3",
            endpoint_url=os.environ.get(
                "S3_ENDPOINT", "https://s3.us-east-005.backblazeb2.com"
            ),
            aws_access_key_id=os.environ["S3_KEY_ID"],
            aws_secret_access_key=os.environ["S3_APP_KEY"],
        )
    except KeyError:
        logger.error(
            "Error: Please set the environment variables S3_KEY_ID and S3_APP_KEY"
        )
    return s3_resource.Bucket(os.environ.get("S3_BUCKET_NAME", "nerc-invoicing"))
```

If the attempt to reference the `S3_*` environment variables raises a `KeyError`, you should probably *exit* rather than simply logging an error. Currently, you log a a message and then proceed to execute...

```python
    return s3_resource.Bucket(os.environ.get("S3_BUCKET_NAME", "nerc-invoicing"))
```

...which fails because `s3_resource` has not been defined.

process_report/config.py
```diff
@@ -0,0 +1,175 @@
...
+# Defaults to being fetched in ColdfrontFetchProcessor
+COLDFRONT_API_FILEPATH = ""
+
+
+### Miscellaneous config values TODO: Should these have their own getter functions?
+NEW_PI_CREDIT_AMOUNT = rates_info.get_value_at("New PI Credit", INVOICE_MONTH, Decimal)
+LIMIT_NEW_PI_CREDIT_TO_PARTNERS = rates_info.get_value_at(
+    "Limit New PI Credit to MGHPCC Partners", INVOICE_MONTH, bool
+)
+BU_SUBSIDY_AMOUNT = rates_info.get_value_at("BU Subsidy", INVOICE_MONTH, Decimal)
```
QuanMPhm, line 79: @larsks @knikolla My only question is should these values that are fetched from `nerc-rates` get their own getter functions? Having this code ran on import means local testing can't be done offline. Aside from that, the code that use these values can have them injected at runtime, so dependency injection during unit tests works fine.

knikolla:
> @larsks @knikolla I've decided to add `getter` functions for configurations that require fetching from S3. This PR is now ready for review

@QuanMPhm you are trying to treat everything as being the same type of "configuration", however you have different types of things which may warrant being treated with different mechanisms
- strings that define paths in s3
- numbers that defines values that are either fetched from the CLI or nerc-rates
- lists
- old pis
- aliases

Some can be generalized, like the first two. While others may require a specialized solution by implementing functions in either the Invoice base class or somewhere else.

process_report/config.py
QuanMPhm: By getter functions, I'm referring to functions like:
```python
@functools.lru_cache
def get_old_pi_filepath() -> str:
    return OLD_PI_FILEPATH or util.fetch_s3(PI_S3_FILEPATH)
```
...in contrast to what my original draft proposed:
```python
OLD_PI_FILEPATH = util.fetch_s3(PI_S3_FILEPATH)
```

The problem with the original draft was that `config.py` would try to fetch from S3 on import, making testing and dependency injection a lot more difficult. Making a "getter" function meant the fetch S3 code can be ran on-demand (or at runtime), allowing dependency injection.

With the getter function, processors/invoices can access configurations like this:
```python
old_pi_filepath: str = field(default_factory=config.get_old_pi_filepath)
```
And since it's just an argument in their `__init__`, during unit tests, I can inject some other value as I wish

QuanMPhm:
@larsks @knikolla I have decided to use `Pydantic` to validate configurations.

process_report/config.py
```diff
@@ -0,0 +1,175 @@
...
+        for line in f:
+            pi_alias_info = line.strip().split(",")
+            alias_dict[pi_alias_info[0]] = pi_alias_info[1:]
+
+    return alias_dict
+
+
+@functools.lru_cache
+def get_prepaid_credits_df() -> pandas.DataFrame:
+    pandas.read_csv(PREPAY_CREDITS_FILEPATH)
```
larsks, line 155: This function does not return anything. All of the tests are passing, which suggests that this function isn't tested. In fact, a test coverage report shows that most of the functions in this file are not tested.

QuanMPhm:
@larsks Apologies. I forgot to actually push up my latest changes that use `Pydantic`.

> This function does not return anything. All of the tests are passing, which suggests that this function isn't tested. In fact, a test coverage report shows that most of the functions in this file are not tested.

I will submit unit tests for this soon :P

process_report/config.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def get_prepaid_projects_df(self) -> pandas.DataFrame:
+        return pandas.read_csv(self.PREPAY_PROJECTS_FILEPATH)
+
+    def get_prepaid_contacts_df(self) -> pandas.DataFrame:
+        return pandas.read_csv(self.PREPAY_CONTACTS_FILEPATH)
+
+
+# Custom configurations goes here
+CONFIG_DICT = {}
```
larsks, line 183: What is `CONFIG_DICT` for?

process_report/config.py
```diff
@@ -0,0 +1,184 @@
...
+        return self.CSV_INVOICE_FILEPATH_LIST
+
+    def get_prepay_debits_filepath(self) -> pydantic.FilePath:
+        if not self.PREPAY_DEBITS_FILEPATH:
+            self.PREPAY_DEBITS_FILEPATH = util.fetch_s3(self.PREPAY_DEBITS_S3_FILEPATH)
+        return self.PREPAY_DEBITS_FILEPATH
+
+    def get_old_pi_filepath(self) -> pydantic.FilePath:
+        if not self.OLD_PI_FILEPATH:
+            self.OLD_PI_FILEPATH = util.fetch_s3(self.PI_S3_FILEPATH)
```
larsks, line 95: Should we have better error handling around s3 access (e.g., in the case of incorrect auth, connection timeouts, etc)?

process_report/processors/validate_pi_alias_processor.py
```diff
@@ -1,12 +1,13 @@
...
 
+from process_report.config import config
 from process_report.invoices import invoice
 from process_report.processors import processor
 
 
 @dataclass
 class ValidatePIAliasProcessor(processor.Processor):
-    alias_map: dict
+    alias_map: dict = field(default_factory=config.get_alias_map)
```
larsks, line 10: Generally, we expect to see type arguments, like `dict[str, str]`, rather than an untyped `dict`.

process_report/config.py
```diff
@@ -0,0 +1,184 @@
...
+            dataframe["End Date"], format="%Y-%m"
+        )
+
+        mask = (dataframe["Start Date"] <= self.INVOICE_MONTH) & (
+            self.INVOICE_MONTH <= dataframe["End Date"]
+        )
+        return dataframe[mask]["Project"].to_list()
+
+    def get_alias_map(self) -> dict:
+        alias_dict = dict()
```
larsks, line 164: Prefer:

```
alias_dict = {}
```

process_report/config.py
```diff
@@ -0,0 +1,184 @@
...
+        dataframe["End Date"] = pandas.to_datetime(
+            dataframe["End Date"], format="%Y-%m"
+        )
+
+        mask = (dataframe["Start Date"] <= self.INVOICE_MONTH) & (
+            self.INVOICE_MONTH <= dataframe["End Date"]
+        )
+        return dataframe[mask]["Project"].to_list()
+
+    def get_alias_map(self) -> dict:
```
larsks, line 163: Generally, we expect to see type arguments, like `dict[str, str]`, rather than an untyped `dict`.

larsks (changes requested):

larsks:
It's not part of this PR, but note that the environment variables for holding S3 (or other AWS-service) credentials are typically `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`, rather than `S3_KEY_ID` and `S3_APP_KEY`, and for specifying an S3 endpoint `AWS_ENDPOINT_URL_S3`.

QuanMPhm:
Note-to-self. Check @larsks's feedback from this repo: https://github.com/larsks/invoicing-example/tree/main

QuanMPhm:
Note-to-self. A new revision is made on branch `213/config_final`. The old revision is on `213/config_consolidate_pydantic`

process_report/invoices/billable_invoice.py
```diff
@@ -45,20 +39,3 @@ def _prepare_export(self):
         self.export_data = self.data[
             self.data[invoice.IS_BILLABLE_FIELD] & ~self.data[invoice.MISSING_PI_FIELD]
         ]
-        self.updated_old_pi_df = self.updated_old_pi_df.astype(
```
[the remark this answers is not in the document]
QuanMPhm: I decided the code to export the Old PI file should be placed in `NewPICreditProcessor` instead. This would remove the need for `BillableInvoice` to contain code that loads in configurations.

Should I have this change be another PR instead?

PR #218, opened 2025-07-09
Closes #215

### LINKED ISSUE #215: Implement basic E2E test for invoicing pipeline validation
@knikolla @QuanMPhm @naved001 discussed yesterday that it would be useful to implement an end-to-end (E2E) test to validate that the invoicing pipeline runs successfully from start to finish. This test is not intended to cover all edge cases or lower-level logic (which are already covered by existing unit test), but instead to ensure that:

- Key pipeline stages are executed successfully in sequence
- No unexpected errors or exceptions occur during execution
- Overall pipeline outputs are structurally correct (sanity check)

**Requirements:** 

- Trigger the full pipeline as it would run in production
- Check basic outputs (do critical files and artifacts exist, are they correct structurally?)
- Did the pipeline execute successfully

**Out of Scope:**

- Fine-grained validation of intermediate data or output contents
- Extensive test coverage of every code path and edge cases 

Deliverables:

- A test script or suite for the e2e test
- Documentation on how to trigger the test locally and/or in CI/CD
- Add to CI pipeline as an optional stage for now

jimmysway:
Implemented E2E testing, added synthetic test data, added basic output validation

added pytest to requirements.txt

Added synthetic test data for the e2e test


process_report/tests/e2e/test_e2e_pipeline.py
```diff
@@ -0,0 +1,191 @@
...
+    """Set up the workspace by copying test data and pipeline code."""
+    # Copy test data files
+    test_files = {}
+    for test_file in test_data_dir.glob("*"):
+        dest_path = workspace / test_file.name
+        shutil.copy(test_file, dest_path)
+        test_files[test_file.name] = dest_path
+
+    # Copy pipeline code
+    process_report_src = project_root / "process_report"
```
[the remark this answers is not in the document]
jimmysway: Oh right, when I was testing, every time I would run the e2e it would default to dumping the files into the root directory, I thought this behaviour wouldn't be ideal because then it would cause potential problems when locally testing and actually running the pipeline vs running the e2e test.
 
I also did not want to change code beyond the realm of the testing, and it couldn't overwrite the location that the test files would end up in, so this solution of copying the code in the temp workspace was the solution I came up with, although intuitively it feels wrong.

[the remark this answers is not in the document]
knikolla: You don't need to copy any of the files.

- You can set the `PYTHONPATH` before running the tests and Python will be able to find the package. - You can also change directory and that will make it so the output files are generated in the directory which you are running the command from.
- You should be able to pass an absolute path for the test data so that it is loaded from the files in the test data directory.

process_report/tests/e2e/test_e2e_pipeline.py
```diff
@@ -0,0 +1,230 @@
...
+        workspace: Path to the temporary workspace directory where the pipeline will run.
+
+    Returns:
+        subprocess.CompletedProcess: The result of the pipeline execution.
+
+    Raises:
+        pytest.fail: If the pipeline execution times out.
+    """
+    logger = logging.getLogger(__name__)
+    logger.info(f"Running pipeline in: {workspace}")
```
QuanMPhm, line 150: Why didn't you create the logger at the top-level of this file instead?

QuanMPhm (changes requested):

process_report/tests/e2e/test_e2e_pipeline.py
```diff
@@ -124,9 +126,11 @@ def _prepare_pipeline_execution(
         str(test_files["test_prepay_contacts.csv"]),
     ]
 
-    # Set up environment
+    # Environment setup for subprocess execution
     env = os.environ.copy()
```
QuanMPhm, line 130: Is there a reason you copied `os.environ`?

process_report/tests/e2e/test_e2e_pipeline.py
```diff
@@ -124,9 +126,11 @@ def _prepare_pipeline_execution(
...
-    env["CHROME_BIN_PATH"] = CHROME_BIN_PATH
+    # Fallback ensures test works even when CI environment doesn't set Chrome path
+    if "CHROME_BIN_PATH" not in env:
+        env["CHROME_BIN_PATH"] = "/usr/bin/chromium"
```
QuanMPhm, line 133: You could instead do this:
```suggestion
env.setdefault("CHROME_BIN_PATH", "/usr/bin/chromium")
```

process_report/tests/e2e/test_e2e_pipeline.py
```diff
@@ -178,24 +181,35 @@ def _validate_outputs(workspace: Path) -> None:
         AssertionError: If any expected output file is missing, empty, or invalid.
         pytest.fail: If CSV files cannot be read or parsed.
     """
-    logger = logging.getLogger(__name__)
     logger.info(f"Validating pipeline outputs in: {workspace}")
 
-    # Validate CSV files
+    # Basic existence and structure validation catches pipeline execution failures
```
QuanMPhm, line 186: I would remove this comment. I would say your validation code here has explained itself.

process_report/tests/e2e/test_e2e_pipeline.py
```diff
@@ -178,24 +181,35 @@ def _validate_outputs(workspace: Path) -> None:
...
+    actual_csv_files = set(csv_path.name for csv_path in workspace.glob("*.csv"))
+    expected_csv_files = set(EXPECTED_CSV_FILES)
+
+    unexpected_files = actual_csv_files - expected_csv_files
+    if unexpected_files:
+        pytest.fail(f"Unexpected CSV files created: {sorted(unexpected_files)}")
+
+    missing_files = expected_csv_files - actual_csv_files
+    if missing_files:
+        pytest.fail(f"Expected CSV files missing: {sorted(missing_files)}")
```
QuanMPhm, line 210: This could be collasped into one test:
```suggestion
assert not (actual_csv_files - expected_csv_files), f"Output dir contains unexpected or missing CSVs: {actual_csv_files}"
```

QuanMPhm (changes requested):

process_report/tests/e2e/test_e2e_pipeline.py
QuanMPhm: @naved001 I was thinking that which files were missing/unexpected could be implied from inspecting `actual_csv_files`, but I guess having it as it is is fine as well.

process_report/tests/e2e/test_e2e_pipeline.py
jimmysway: I thought by copying the environment would be defensive in the sense that the pattern is: Copy current environment > modify the copy > pass to subprocess. So surely copying os.environ will give us an isolated environment to modify without affecting the current process's global environment variables.

process_report/tests/e2e/README.md
```diff
@@ -0,0 +1,121 @@
...
+python -m unittest discover -s process_report/tests/e2e
+
+# Or run the specific test file
+python -m unittest process_report.tests.e2e.test_e2e_pipeline
+
+# Or run directly
+python process_report/tests/e2e/test_e2e_pipeline.py
+```
+
+### Environment Variables
```
naved001, line 63: This section should be moved before the local execution section because we would want to set the variables before running the tests.

process_report/tests/e2e/README.md
```diff
@@ -0,0 +1,121 @@
...
+### Local Execution
+
+Run the E2E tests locally:
+
+```bash
+# From the project root directory
+python -m unittest discover -s process_report/tests/e2e
+
+# Or run the specific test file
+python -m unittest process_report.tests.e2e.test_e2e_pipeline
```
naved001, line 57: These 2 command didn't run the tests. `test_e2e_pipeline.py` does not use the unittest module. The tests need to be in a class that inherit from the base unittest.TestCase. 

You should just document that you expect this to run with pytest (as you are doing it in the github workflows) and don't worry about unittest at all.

process_report/tests/e2e/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+# Or run directly
+python process_report/tests/e2e/test_e2e_pipeline.py
```
naved001, line 60: This also does not work because there's nothing in `test_e2e_pipeline.py` that actually calls the test method

process_report/tests/e2e/README.md
```diff
@@ -0,0 +1,121 @@
...
+- **Configuration Files**: Non-billable PIs/projects, timed projects, prepay configurations
+- **Historical Data**: Old PI file, alias mappings, prepay debits
+
+## Running the Tests
+
+### Prerequisites
+
+1. Install dependencies:
+   ```bash
+   pip install -r requirements.txt
```
naved001, line 36: specify installing the test requirements as well.

.github/workflows/e2e-tests.yaml
```diff
@@ -0,0 +1,34 @@
...
+        sudo apt-get install -y chromium
+
+    - name: Install dependencies
+      run: |
+        pip install -r requirements.txt
+        pip install -r process_report/tests/test-requirements.txt
+
+    - name: Run E2E Tests
+      run: |
+        pytest process_report/tests/e2e/test_e2e_pipeline.py -v
```
naved001, line 32: Just a suggestion, in my opinion we should run this with `-s` flag so we can see the output when the tests run.

process_report/tests/e2e/test_e2e_pipeline.py
```diff
@@ -0,0 +1,241 @@
...
+            command,
+            env=env,
+            cwd=workspace,
+            capture_output=True,
+            text=True,
+            timeout=PIPELINE_TIMEOUT,
+        )
+
+        if result.stderr:
+            logger.warning(f"Pipeline stderr: {result.stderr}")
```
naved001, line 164: Currently, there's some stuff that gets written to stderr but I don't know where the output of this logger is going. I did run pytest with `-s` to disable all capturing but I still did not see this specific warning or anything else that's being logged. 

But anyway, when I do run pytest with `-s` then I see this result anyway because it then prints the stderr.

naved001 (dismissed):
Thanks @jimmysway for putting this together. Really appreciate the dosctrings everywhere!

QuanMPhm (changes requested):
Usually, for a PR as substantive as this one, I would prefer a more in-depth summary of changes. An example would be:

```suggestion
Closes #215. Implemented End-to-end (e2e) testing

A new workflow has been which runs the e2e test. This test performs 
validation of the pipeline’s output given a minimal input test data.
Aded a new `test-requirements.txt` file that contains `pytest`.
```

QuanMPhm (approved):

PR #219, opened 2025-07-15
Chromium test is not a unit test

larsks:
Relocate tests/unit/test_chromium.py to tests/integration/test_chromium.py.

The chromium test is not a unit test -- it has an external dependency on
Chromium. By relocating it to tests/integration, we ensure that the unit
tests run without errors.

This commit also contains some minor changes to suppress warnings from
strict type checkers.

### DIFF
--- a/process_report/invoices/pi_specific_invoice.py
+++ b/process_report/invoices/pi_specific_invoice.py
@@ -13,6 +13,7 @@ import process_report.util as util
 
 
 TEMPLATE_DIR_PATH = "process_report/templates"
+CHROME_BIN_PATH = os.environ.get("CHROME_BIN_PATH", "/usr/bin/chromium")
 
 
 logger = logging.getLogger(__name__)
@@ -120,12 +121,9 @@ class PIInvoice(invoice.Invoice):
             temp_fd.flush()
 
         def _create_pdf_invoice(temp_fd_name):
-            chrome_binary_location = os.environ.get(
-                "CHROME_BIN_PATH", "/usr/bin/chromium"
-            )
-            if not os.path.exists(chrome_binary_location):
+            if not os.path.exists(CHROME_BIN_PATH):
                 sys.exit(
-                    f"Chrome binary does not exist at {chrome_binary_location}. Make sure the env var CHROME_BIN_PATH is set correctly and that Google Chrome is installed"
+                    f"Chrome binary does not exist at {CHROME_BIN_PATH}. Make sure the env var CHROME_BIN_PATH is set correctly and that Google Chrome is installed"
                 )
 
             invoice_pdf_path = (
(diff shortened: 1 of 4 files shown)


KelvinLinBU (approved):
Looks good!

process_report/tests/unit/invoices/test_pi_specific_invoice.py
```diff
@@ -137,10 +138,10 @@ def test_export_pi(self, mock_subprocess_run, mock_path_exists, mock_filter_cols
...
                 answer_arglist = [
-                    "/usr/bin/chromium",
+                    CHROME_BINARY_LOCATION,
                     "--headless",
                     "--no-sandbox",
                     f"--print-to-pdf={pi_pdf_path}",
                     "--no-pdf-header-footer",
                 ]
-                self.assertTrue(answer_arglist == chrome_arglist[0][:-1])
+                assert answer_arglist == chrome_arglist[0][:-1]
```
QuanMPhm, line 147: Is there a benefit to using `assert` instead of `self.assertTrue`? My initially feeling is that we should just stick with `self.assertTrue` since that's what every other test in this repo does.

KelvinLinBU: I'll create an issue and see if we can update it throughout the repo.

QuanMPhm (approved):

QuanMPhm:
@naved001 @knikolla Just need a pass from either of you

process_report/invoices/pi_specific_invoice.py
```diff
@@ -120,20 +121,17 @@ def _create_html_invoice(temp_fd):
...
         def _create_pdf_invoice(temp_fd_name):
-            chrome_binary_location = os.environ.get(
-                "CHROME_BIN_PATH", "/usr/bin/chromium"
-            )
-            if not os.path.exists(chrome_binary_location):
+            if not os.path.exists(CHROME_BINARY_LOCATION):
                 sys.exit(
-                    f"Chrome binary does not exist at {chrome_binary_location}. Make sure the env var CHROME_BIN_PATH is set correctly and that Google Chrome is installed"
+                    f"Chrome binary does not exist at {CHROME_BINARY_LOCATION}. Make sure the env var CHROME_BIN_PATH is set correctly and that Google Chrome is installed"
                 )
```
naved001, line 127: it'd be nice if we called the python constant and the env variable the same - CHROME_BINARY_LOCATION vs CHROME_BIN_PATH.

naved001 (approved):

process_report/invoices/pi_specific_invoice.py
larsks: That seems reasonable.

PR #225, opened 2025-08-11
Update institute_list.yaml

joachimweyl:


process_report/institute_list.yaml
```diff
@@ -111,3 +111,7 @@
 -   display_name: AltaStata
     domains:
     - altastata.com
+-   display_name: The Synergist
```
naved001, line 114: @joachimweyl The tests are failing because this name already exsists for the domain "thesynergist.org". I can't comment on the line but it's line 101

naved001 (changes requested):

PR #227, opened 2025-08-12
Replace unittest assert with pytest asserts

### LINKED ISSUE #220: Replace unittest assertions with assert statements for pytest compatibility
Refactor unittest assertion methods (e.g. self.assertTrue) to use plain assert statements for improved readability and better failure output when using pytest. This simplifies the test code and aligns it with pytest-style testing.

KelvinLinBU:
Closes #220. Update test workflows to use `pytest`

As suggested before[1], `unittest` assertions are now replaced 
with normal asserts to utilize  `pytest`'s better error logging. 
Corresponding changes made to workflows.

[1] https://github.com/CCI-MOC/invoicing/issues/220#issuecomment-3074211605

### DIFF
--- a/.github/workflows/integration-tests.yaml
+++ b/.github/workflows/integration-tests.yaml
@@ -23,6 +23,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -r requirements.txt
+          pip install -r process_report/tests/test-requirements.txt
 
       - name: Download and install latest Chromium
         run: |
@@ -31,4 +32,4 @@ jobs:
 
       - name: Run integration tests
         run: |
-          python -m unittest discover -s process_report/tests/integration
+          python -m pytest process_report/tests/integration
--- a/.github/workflows/unit-tests.yaml
+++ b/.github/workflows/unit-tests.yaml
@@ -19,7 +19,8 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -r requirements.txt
+          pip install -r process_report/tests/test-requirements.txt
 
       - name: Run unit tests
         run: |
-          python -m unittest discover -s process_report/tests/unit
+          python -m pytest process_report/tests/unit
--- a/process_report/tests/integration/test_chromium.py
+++ b/process_report/tests/integration/test_chromium.py
@@ -54,11 +54,11 @@ class TestPIInvoiceExport(BaseTestCaseWithTempDir):
 
         output_pdfs = os.listdir(self.tempdir)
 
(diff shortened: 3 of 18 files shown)


.github/workflows/integration-tests.yaml
```diff
@@ -23,6 +23,7 @@ jobs:
       - name: Install dependencies
         run: |
           pip install -r requirements.txt
+          pip install pytest
```
QuanMPhm, line 26: In many of our repos, we put the dependancies for testing into a `test-requirements.txt` file. Please create this file and add `pytest` into it, and have `pip install -r test-requirements.txt` here instead

.github/workflows/integration-tests.yaml
```diff
@@ -31,4 +32,4 @@ jobs:
 
       - name: Run integration tests
         run: |
-          python -m unittest discover -s process_report/tests/integration
+          python -m pytest -q process_report/tests/integration
```
QuanMPhm, line 35: I would remove the `-q` flag here, since I think it's better to keep normal verbosity.

QuanMPhm (changes requested):
Looks good. Such a minor comment

QuanMPhm (changes requested):
I'll also say that the 2 commits in this PR should be combined into one. The first commit would not work without the second commit, since it requires `pytest` and would throw an error because pytest is not installed yet. Each of your commits should be standalone. They should work and pass all the tests by themselves. Having commits that depend on one another makes rollbacks or rebasing very messy.

QuanMPhm (changes requested):
The code changes looks good. But could you update the commit message? It's not accurate to say you're closing a comment. Lars' comment also didn't mention changing any workflows. Your workflow changes are a indirect requirement of using `pytest`. My suggestion for the commit message would be:
```suggestion
Update test workflows to use `pytest`

As suggested before[1], `unittest` assertions are now replaced 
with normal asserts to utilize  `pytest`'s better error logging. 
Corresponding changes made to workflows.

[1] https://github.com/CCI-MOC/invoicing/issues/220#issuecomment-3074211605
```

Also, please limit the line length of your commit messages to [72 characters](https://dev.to/noelworden/improving-your-commit-message-with-the-50-72-rule-3g79) (as in my example above). Any longer and it may become harder to read on certain terminals.

QuanMPhm (approved):

QuanMPhm (approved):
@larsks Your campaign to have us use `pytest` is coming to fruition

PR #231, opened 2025-08-19
Added fetch function as an out-of-band utility to easily fetch invoices

### LINKED ISSUE #92: Implement a `fetch` function on the Invoice
One of the steps that Kim has to go through is generating graphs and insights from invoices of previous months. We can easily automate this by fetching the already uploaded invoices based on month and running whatever comparisons we need.

We should therefore be able to do

```
b_08 = BillableInvoice(invoice_month="2024-08")
b_08.fetch(s3_bucket)
b_07 = BillableInvoice(invoice_month="2024-07")
b_07.fetch(s3_bucket)
```

And this would give us the invoice as exported for 2024-08 and 2024-07 in the `data` dataframe of the objects `b_`. We know where to download from s3 if #91 is implemented. 

Eventually, we only really want the singular dataframe (after all processors) to be fetchable (which we will also export as is) as applying the future "views" will get us the invoices we need exactly. (Do not worry about PI invoices)

jimmysway:
Closes #92 

Added fetch function as an out-of-band utility to easily fetch invoices

### DIFF
--- a/process_report/invoices/invoice.py
+++ b/process_report/invoices/invoice.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
 import pandas
+import os
 
 import process_report.util as util
 
@@ -122,3 +123,20 @@ class Invoice:
     def export_s3(self, s3_bucket):
         s3_bucket.upload_file(self.output_path, self.output_s3_key)
         s3_bucket.upload_file(self.output_path, self.output_s3_archive_key)
+
+    def fetch(self, s3_bucket=None):
+        """Fetches the exported invoice CSV from S3 and loads it into self.data.
+
+        If s3_bucket is not provided, uses the default invoice bucket.
+        The CSV is downloaded to the current working directory using its base filename.
+        """
+        s3_key = self.output_s3_key
+        local_filename = os.path.basename(s3_key)
+
+        if s3_bucket is None:
+            # Fallback to default bucket utility which also downloads
+            util.fetch_s3(s3_key)
+        else:
+            s3_bucket.download_file(s3_key, local_filename)
+
+        self.data = pandas.read_csv(local_filename)


process_report/tests/unit/invoices/test_invoice_fetch.py
QuanMPhm, line 1: I would suggest that you add this test into `test_base_invoice.py` instead since `fetch()` is a function belonging to the base `Invoice` class. Doing so will also mean the test will be discovered by `unittest`. Right now, this file cannot be discovered by `unittest`, and so your test is never ran by the workflow.

process_report/tests/unit/invoices/test_invoice_fetch.py
```diff
@@ -0,0 +1,35 @@
+import pandas as pd
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from pytest_mock.plugin import MockerFixture
+
+from process_report.tests import util as test_utils
+
+
+def test_fetch_with_mock_s3_bucket(mocker: "MockerFixture") -> None:
```
QuanMPhm, line 10: Regardless if the test is compatible for `unittest` or `pytest`, I would suggest you use `unittest.mock` to mock the S3 bucket and `read_csv`, since that is what we use in the rest of this repo. This will also avoid the need to add another dependancy

process_report/tests/unit/invoices/test_invoice_fetch.py
```diff
@@ -0,0 +1,35 @@
...
+    )
+
+    mocker.patch("pandas.read_csv", return_value=expected_data)
+    mock_s3_bucket = mocker.Mock()
+    mock_s3_bucket.download_file.return_value = None
+
+    test_invoice.fetch(mock_s3_bucket)
+
+    assert test_invoice.data is not None
+    pd.testing.assert_frame_equal(test_invoice.data, expected_data)
```
QuanMPhm, line 32: I would suggest 
```suggestion
    assert test_invoice.data.equals(expected_data)
```
I think this looks simpler, and is also what the rest of the repo does

PR #240, opened 2025-10-21
Projects that belong to courses are now non-billable

QuanMPhm:
Closes https://github.com/CCI-MOC/non-billable-projects/issues/66. Leveraging the private field `Is Course?` in Coldfront [1], if this field is set to `Yes`, the project/allocation is considered part of a course and therefore non-billable.
If the field is not set, it's assumed to be not part of a course

[1] https://github.com/ubccr/coldfront/blob/a4b6b4c1d18535f61de5aa4bbfe509d7d8ca4b0f/coldfront/core/allocation/management/commands/add_allocation_defaults.py#L56


joachimweyl:
@QuanMPhm do we know what the conflicts are that need to be resolved?

PR #262, opened 2026-02-12
Fixed Pandas read_csv bug by specifying engine

QuanMPhm:
As mentioned by Kristi[1], the better solution to the Pandas read_csv bug is to specify the engine as "pyarrow", rather than having the loading and casting step seperate.

[1] https://github.com/nerc-project/coldfront-plugin-cloud/pull/290#pullrequestreview-3787503960


knikolla (approved):
As mentioned in the other PR, I think it's worth updating the dependency to require pandas `>=3` and `<4`.

PR #267, opened 2026-02-17
Added nerc_invoicing repo with functions for querying invoicing Iceberg table

### LINKED ISSUE #266: Query lifetime cost of a specific (Allocation Name, Cluster Name)
## Motivation
Phase 3 of [Invoicing](https://docs.google.com/document/d/1Qz1b3ZtWCundt7IYOPB-mP8Uha6qSKUJtNnrNdxtO6c/edit?tab=t.0#heading=h.1uv8tt5glsr) rework requires support for alerting PIs when a cost value for the lifetime of an (allocation name, cluster name) has been reached. This requires exposing a function that allows querying the Iceberg table with the historical costs of a allocations and running a sum for a given allocation. 

## Completion Criteria
- Iceberg table is accessible by external clients such as ColdFront
- Function implemented in `invoicing` that can be used by clients to query lifetime cost of a specific allocation + cluster name.

## Description
- [ ] 

## Completion dates
Desired - 2026-02-27
Required - TBD

jimmysway:
Added nerc_invoicing repo for querying invoicing Iceberg table

Added functionality to get lifetime costs grouped by project

Added pyproject.toml for future publishing

Closes #266

### DIFF
--- a/process_report/data_tools/config.py
+++ b/process_report/data_tools/config.py
@@ -1,7 +1,4 @@
-import functools
-
 from pydantic_settings import BaseSettings
-from pyiceberg.table import StaticTable
 
 
 class DataToolsSettings(BaseSettings):
@@ -39,11 +36,3 @@ class DataToolsSettings(BaseSettings):
 
 
 data_tools_settings = DataToolsSettings()
-
-
-@functools.cache
-def get_table() -> StaticTable:
-    return StaticTable.from_metadata(
-        data_tools_settings.table_path,
-        properties=data_tools_settings.iceberg_s3_properties(),
-    )
--- a/process_report/data_tools/config.py
+++ b/process_report/data_tools/config.py
@@ -1,0 +1,49 @@
+import functools
+
+from pydantic_settings import BaseSettings
+from pyiceberg.table import StaticTable
+
+
+class DataToolsSettings(BaseSettings):
+    """Iceberg warehouse path and S3 credentials for data_tools queries."""
+
+    iceberg_warehouse_base: str = "s3://nerc-invoicing-iceberg/warehouse"
+    iceberg_table_subpath: str = "nerc_invoicing_iceberg/nerc_invoicing_iceberg"
+    iceberg_s3_access_key_id: str | None = None
+    iceberg_s3_secret_access_key: str | None = None
+    iceberg_s3_endpoint: str | None = None
+    iceberg_s3_region: str = "us-east-005"
+
+    @property
+    def table_path(self) -> str:
+        return f"{self.iceberg_warehouse_base}/{self.iceberg_table_subpath}"
+
+    def iceberg_s3_properties(self) -> dict[str, str]:
+        if not all(
+            [
+                self.iceberg_s3_access_key_id,
+                self.iceberg_s3_secret_access_key,
+                self.iceberg_s3_endpoint,
+            ]
+        ):
+            raise ValueError(
+                "Iceberg S3 credentials required: "
+                "ICEBERG_S3_ACCESS_KEY_ID, ICEBERG_S3_SECRET_ACCESS_KEY, ICEBERG_S3_ENDPOINT"
+            )
+        return {
+            "s3.access-key-id": self.iceberg_s3_access_key_id,
+            "s3.secret-access-key": self.iceberg_s3_secret_access_key,
+            "s3.endpoint": self.iceberg_s3_endpoint,
+            "s3.region": self.iceberg_s3_region,
+        }
+
+
+data_tools_settings = DataToolsSettings()
+
+
+@functools.cache
+def get_table() -> StaticTable:
+    return StaticTable.from_metadata(
+        data_tools_settings.table_path,
+        properties=data_tools_settings.iceberg_s3_properties(),
+    )
--- a/process_report/data_tools/costs.py
+++ b/process_report/data_tools/costs.py
@@ -1,0 +1,162 @@
+import functools
+from decimal import Decimal
+import logging
+
+import pandas as pd
+from pyiceberg.expressions import And, BooleanExpression, EqualTo
+
+import process_report.invoices.invoice as invoice
+from process_report.data_tools.config import get_table
+
+logger = logging.getLogger(__name__)
+FilterValue = str | int | float
+
+_LIFETIME_COLS = [
+    invoice.PROJECT_ID_FIELD,
+    invoice.CLUSTER_NAME_FIELD,
+    invoice.BALANCE_FIELD,
+]
+
+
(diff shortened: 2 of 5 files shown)


nerc_invoicing/costs.py
```diff
@@ -0,0 +1,35 @@
+import pandas as pd
+
+import config
```
knikolla: Always import things as either
- Fully qualified modules (i.e. `invoicing.config`)
- From the same folder (i.e. `.config`)

knikolla (changes requested):

data_tools/costs.py
```diff
@@ -0,0 +1,118 @@
...
+    df[agg_col] = df[agg_col].fillna(0)
+    agg_spec = {agg_name: (agg_col, "sum")}
+    grouped_df = df.groupby(list(group_by), as_index=False).agg(**agg_spec)
+    grouped_df[agg_name] = grouped_df[agg_name].map(
+        lambda v: Decimal(str(v)).quantize(Decimal("0.01"))
+    )
+    return grouped_df
+
+
+def calculate_lifetime_costs(**filters: str | int | float) -> pd.DataFrame:
```
knikolla: Give me an example of `filters` that would be used with this function.

knikolla (changes requested):

knikolla (changes requested):
Previous comments still unresolved.

data_tools/costs.py
jimmysway: For example if you just wanted the costs for `vllm-test`

You could run:

```
  filters = {
      invoice.PROJECT_ID_FIELD: "vllm-test",
  }
  print(calculate_lifetime_costs(**filters))
```
   
I will add an example to the docstring or update a readme in the future

knikolla (changes requested):
Comments about using the config system and caching still applicable.

process_report/data_tools/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .costs import calculate_lifetime_costs
```
knikolla: Why not leave this file as empty?

knikolla (changes requested):

jimmysway:
> > > Comments about using the config system and caching still applicable.
> > 
> > 
> > Sorry I missed those I am addressing those now. I think caching the `get_invoice_dataframe()` function might be premature. The function returns a dataframe potentially for a specific filter if we are only querying this dataframe once there is little benefit from caching right now.
> > I will instead apply it to the `_get_table()` function
> 
> Why do you think `_get_table()` is a better place for implementing caching?

I applied caching to _get_table() instead because it returns a reusable StaticTable handle (metadata + query interface), not full row data. Caching _get_table() mainly reduces repeated table-initialization overhead on every call. 

If we were instead to cache the dataframe itself with `get_invoice_dataframe()`. We would instead be caching filter-specific dataframes which we don't reuse much right now. 

I guess the question is what is the purpose of this library? Are we going to use as a thin wrapper for downstream processes to run predictable queries once a month, or are we going to be using this as a primary way to interface with Iceberg where it will be used as other processes or data tools to interface and run analytics on the data itself. 

Currently, I think the library is going to be used as a thin wrapper. Since this module is just a thin wrapper, it should do the simple safe optimization (_get_table) and leave heavier caching to whatever service or job is using it.

process_report/data_tools/config.py
```diff
@@ -0,0 +1,49 @@
...
+            ]
+        ):
+            raise ValueError(
+                "Iceberg S3 credentials required: "
+                "ICEBERG_S3_ACCESS_KEY_ID, ICEBERG_S3_SECRET_ACCESS_KEY, ICEBERG_S3_ENDPOINT"
+            )
+        return {
+            "s3.access-key-id": self.iceberg_s3_access_key_id,
+            "s3.secret-access-key": self.iceberg_s3_secret_access_key,
+            "s3.endpoint": f"https://{self.iceberg_s3_endpoint}",
```
QuanMPhm: For coherence with how we're handling the S3 URL in this and other repos, could you remove the `https://` from the format string? In fact, just have it be:
```
"s3.endpoint": self.iceberg_s3_endpoint,
```

QuanMPhm (changes requested):
Some comments I have for now. Will take a closer look and run some tests

process_report/data_tools/costs.py
```diff
@@ -0,0 +1,162 @@
...
+    if row_filter:
+        scan = table.scan(row_filter=row_filter)
+    else:
+        scan = table.scan()
+    if cols:
+        scan = scan.select(*cols)
+    df = scan.to_pandas()
+    if filters and df.empty:
+        logger.warning("No invoice rows matched filters: %s", filters)
+    return df
```
QuanMPhm: Follow-up from above comment, would this be simpler?

```suggestion
def get_invoice_dataframe(
    cols: tuple[str, ...] = ("*",), **filters: FilterValue
) -> pd.DataFrame:
    """Load invoice data from the Iceberg table.

    Args:
        cols: Column names to select as a tuple. Defaults to all columns
        **filters: Column names as keys, values to filter by. Values must be str, int, or float.

    Returns:
        DataFrame of invoice data from the table.
    """
    table = get_table()
    row_filter = _row_filter(**filters)
    scan = table.scan(row_filter=row_filter, selected_fields=cols)
    df = scan.to_pandas()
    if filters and df.empty:
        logger.warning("No invoice rows matched filters: %s", filters)
    return df
```

QuanMPhm (changes requested):
Thank you for the good work. Just a few more comments mostly regarding style.

QuanMPhm:
@jimmysway Some CI test cases are broken

process_report/data_tools/costs.py
```diff
@@ -0,0 +1,162 @@
...
+        PyIceberg BooleanExpression like EqualTo(col1, 'x') AND EqualTo(col2, 1),
+        or None if no filters are given.
+    """
+    if not filters:
+        return None
+    expression: BooleanExpression | None = None
+    for col, val in filters.items():
+        clause = EqualTo(col, val)
+        expression = clause if expression is None else And(expression, clause)
+    return expression
```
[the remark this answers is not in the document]
QuanMPhm: It seems passing `None` to `table.scan()` raises an error:
```python
table.scan(None).to_pandas()
...
NotImplementedError: Cannot visit unsupported expression: None

```

I'll follow my suggestion right now with `_row_filter()` only returning `BooleanExpression`

QuanMPhm:
@knikolla @naved001 I made some small amendments to this PR. I have one question [here](https://github.com/CCI-MOC/invoicing/pull/267#discussion_r2913680986)

PR #272, opened 2026-03-03
Fixing quotechar for names with commas

QuanMPhm:
During 2026-02 invoicing, we found an allocation in the storage service invoice with the name: `|HA25 Versatile, Safe and Personalizable Large Language Mo-e2b2bd|`. Because the [storage invoice ](https://github.com/nerc-project/coldfront-plugin-cloud/blob/a9f41dd750d68de6b58f27d2ec1c1254dfbe5f70/src/coldfront_plugin_cloud/management/commands/calculate_storage_gb_hours.py#L50)uses the quote character `|` we want to update the invoicing code to also use it as well.


knikolla (approved):
Looks good.

Issue #274, opened 2026-03-06
Updating invoicing to handle the schema change from `invoicing-private-data`


jimmysway:
### Part 1

We also need to to update invoicing to handle the logic to accept the new `is_billable` flag in the `projects.yaml` that applies to special cases such as `griot-grits-aa488b`

The PR to the schema update for `projects.yaml` is listed here:

CCI-MOC/non-billable-projects#89

### Part 2

In the ongoing invoicing post-processing automation we found the need to add special overrides for special cases such as the case of emre_keskin@harvard and the best way to do that was to update the schema `pi.txt` to `pi.yaml`.

```
- email: emre_keskin@harvard.edu
  non-billed-su-types:
      - Openstack Storage
```

The PR to the schema update for `pi.txt` to `pi.yaml` is listed here:

CCI-MOC/non-billable-projects#87

PR #279, opened 2026-03-10
Add support for non-billed service unit types 

jimmysway:
Add support for non-billed service unit types 
Support non-billed SU types from pi.yaml 
Apply credit code 0005 and zero balances for matching rows

Part 2 of #274


QuanMPhm:
@knikolla @naved001 I have re-written this PR, since it seemed Jimmy's direction was somewhat off.

PR #282, opened 2026-03-21
Allow supplemental API data to be passed to ColdfrontFetchProcessor

### LINKED ISSUE #239: Fill out Project Name and PI column using bm_projects.csv file in nonbillable repo
Related to https://github.com/CCI-MOC/non-billable-projects/issues/67, we want to the `Project - Allocation` and `Manger (PI)` field to be filed by a `bm_projects.csv` file in the nonbillable repo. Red Hat has project names they want to appear on their invoice.

From discussing with @knikolla on Slack, we should have this file be cluster-agnostic, such that it becomes a supplemental source of information for projects not managed by Coldfront.

QuanMPhm:
Closes #239. Related to https://github.com/CCI-MOC/invoicing-private-data/pull/68. This allows passing data for billable projects not currently in Coldfront Namely bare metal projects.

This will finally allow billing of Bare Metal projects

### DIFF
--- a/process_report/processors/coldfront_fetch_processor.py
+++ b/process_report/processors/coldfront_fetch_processor.py
@@ -2,7 +2,7 @@ import os
 import sys
 import functools
 import logging
-import json
+import yaml
 from dataclasses import dataclass, field
 
 import requests
@@ -32,7 +32,7 @@ class ColdfrontFetchProcessor(processor.Processor):
     nonbillable_projects: pandas.DataFrame = field(
         default_factory=loader.get_nonbillable_projects
     )
-    coldfront_data_filepath: str = invoice_settings.coldfront_api_filepath
+    coldfront_data_filepaths: tuple[str] = invoice_settings.coldfront_api_filepaths
 
     @functools.cached_property
     def coldfront_client(self):
@@ -84,18 +84,22 @@ class ColdfrontFetchProcessor(processor.Processor):
         return r.json()
 
     def _get_coldfront_api_data(self):
-        if self.coldfront_data_filepath:
-            logger.info(
-                f"Using Coldfront data from {self.coldfront_data_filepath} instead of API"
-            )
-            with open(self.coldfront_data_filepath, "r") as f:
-                return json.load(f)
-        else:
-            return self._fetch_coldfront_allocation_api()
+        api_data = []
+        for api_data_file in self.coldfront_data_filepaths:
+            logger.info(f"Using Coldfront data from {api_data_file}")
+            with open(api_data_file) as f:
+                api_data += yaml.safe_load(f)
+
+        if invoice_settings.keycloak_client_id:
+            logger.info("Loading Coldfront data from remote server")
+            api_data += self._fetch_coldfront_allocation_api()
+
+        return api_data
 
     def _get_allocation_data(self, coldfront_api_data):
         """Returns a mapping of (project ID, cluster name) tupels to a dict of project name, PI name, and institution code."""
         allocation_data = {}
+        duplicate_projects = []
         for project_dict in coldfront_api_data:
             try:
                 # Allow allocation to not have institute code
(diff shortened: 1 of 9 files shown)


QuanMPhm:
@knikolla @naved001 The PR has been updated to use the new data format for the supplemental API data. It greatly simplified things

process_report/processors/coldfront_fetch_processor.py
```diff
@@ -33,6 +33,9 @@ class ColdfrontFetchProcessor(processor.Processor):
         default_factory=loader.get_nonbillable_projects
     )
     coldfront_data_filepath: str = invoice_settings.coldfront_api_filepath
+    supplement_api_data: list[dict] = field(
```
knikolla: Wouldn't this all now be simpler now if we converted `coldfront_data_filepath` instead to accepting a list and treat both files as the same, removing the need for the `supplement_api_data` being a separate thing? :)

QuanMPhm: I looked a bit, and its a bit more complicated. [Currently](https://github.com/CCI-MOC/invoicing/blob/main/process_report/processors/coldfront_fetch_processor.py#L86-L94), the presence of `coldfront_data_filepath` determines whether `invoicing` will fetch the API data from Coldfront or not. This is useful for local and e2e testing. 

Putting `supplement_api_data` together with `coldfront_data_filepath` means we won't be able to both fetch Coldfront and provide local supplemental data. It seems we would have to do a bit more refactoring for this to go smoothly.

knikolla: We wrote the decision flow so we can change it to do both. 

1. Are the environment variables describing how to access the ColdFront API present? If yes, fetch.
2. Are there `coldfront_data_filepath` files? Load all of them.
3. Merge 1 and 2.

Sure, there is some slight refactoring to get there and I'll let you decide whether you want to do it as part of this PR or a future one.

QuanMPhm: I did the refactoring. The logic now follows your suggestion

process_report/processors/coldfront_fetch_processor.py
```diff
@@ -84,14 +84,17 @@ def _fetch_coldfront_allocation_api(self):
...
-            return self._fetch_coldfront_allocation_api()
+        api_data = []
+        for api_data_file in self.coldfront_data_filepaths:
+            logger.info(f"Using Coldfront data from {api_data_file}")
+            with open(api_data_file) as f:
+                api_data += yaml.safe_load(f)
+
+        if invoice_settings.keycloak_client_id:
+            logger.info("Loading Coldfront data from remote server")
+            api_data += self._fetch_coldfront_allocation_api()
```
knikolla, line 105: This now introduces the possibility of an Allocated Project ID being matched to several instances in the ColdFront data. Can you please write a test that documents the behavior in such a scenario?

process_report/tests/e2e/test_e2e_pipeline.py
```diff
@@ -112,7 +113,14 @@ def _prepare_pipeline_execution(
     # Environment setup for subprocess execution
     env = os.environ.copy()
     env["INVOICE_MONTH"] = INVOICE_MONTH
-    env["COLDFRONT_API_FILEPATH"] = str(test_files["test_coldfront_api_data.json"])
+
+    # pydantic_settings parses complex types as JSON-encoded strings: https://pydantic.dev/docs/validation/latest/concepts/pydantic_settings/#parsing-environment-variable-values
+    env["COLDFRONT_API_FILEPATHS"] = json.dumps(
+        (
+            str(test_files["test_coldfront_api_data.json"]),
```
knikolla, line 120: In the file parsing code, you actually only call `yaml.safe_load` so you don't actually support JSON files anymore. Please reintroduce logic to check the file extension and call the appropriate loading mechanism for JSON or YAML.

knikolla (changes requested):
Two comments, otherwise looks pretty good :)

process_report/processors/coldfront_fetch_processor.py
QuanMPhm: From 2:30 discussion today, I'll keep the behavior as error out and write the test for it

process_report/tests/e2e/test_e2e_pipeline.py
QuanMPhm: When I saw the e2e test pass, I realized `yaml.safe_load()` can load JSON files as well. Do you still want separate code paths for JSON and YAML loading?

```python
json_data = '{"name": "Alice", "age": 30, "city": "New York"}'
parsed_data = yaml.safe_load(json_data)
# parsed_data is {'name': 'Alice', 'age': 30, 'city': 'New York'}
```

PR #285, opened 2026-04-02
Track initialized and used columns in each processor

### LINKED ISSUE #284: Resolve bug with PI-SU processor
## Motivation
During 2026-03 invoicing, a bug was found where the columns initialized by the New-PI credit processor (i.e `PI Balance` column), was being accessed by the PI-SU processor before it was initialized, causing an KeyError.
## Completion Criteria
Fix the bug
## Description
- [ ] Bug is fixed. PR merged

## Completion dates
Desired - 2026-04-27
Required - 2026-04-30

QuanMPhm:
Closes CCI-MOC/invoicing#284. During 2026-03 invoicing, a bug was found where the columns initialized by the New-PI credit processor (i.e `PI Balance` column), was being accessed by the PI-SU processor before it was initialized, causing an KeyError.

More details in the commit message

[1] https://github.com/CCI-MOC/invoicing/pull/279

### DIFF
--- a/process_report/invoices/invoice.py
+++ b/process_report/invoices/invoice.py
@@ -1,7 +1,30 @@
 from dataclasses import dataclass
+from typing import Any, Callable
+from decimal import Decimal
 import pandas
+import pyarrow
+import logging
 
 import process_report.util as util
+
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
+class InvoiceColumn:
+    name: str
+    dtype: Any
+    default_value: Any | None = None
+    default_initializer: Callable[[pandas.DataFrame], pandas.Series] | None = None
+
+
+# Field type definitions
+BALANCE_FIELD_TYPE = pandas.ArrowDtype(pyarrow.decimal128(21, 2))
+RATE_FIELD_TYPE = pandas.ArrowDtype(pyarrow.decimal128(21, 13))
+INTEGER_FIELD_TYPE = pandas.ArrowDtype(pyarrow.int64())
+STRING_FIELD_TYPE = pandas.StringDtype()
+BOOL_FIELD_TYPE = pandas.BooleanDtype()
 
 
 ### PI file field names
@@ -65,11 +88,69 @@ CLUSTER_NAME_FIELD = "Cluster Name"
 IS_COURSE_FIELD = "Is Course"
 ###
 
+### Initialized Column objects
+INVOICE_DATE_COLUMN = InvoiceColumn(name=INVOICE_DATE_FIELD, dtype=STRING_FIELD_TYPE)
+PROJECT_COLUMN = InvoiceColumn(name=PROJECT_FIELD, dtype=STRING_FIELD_TYPE)
+PROJECT_ID_COLUMN = InvoiceColumn(name=PROJECT_ID_FIELD, dtype=STRING_FIELD_TYPE)
+PI_COLUMN = InvoiceColumn(name=PI_FIELD, dtype=STRING_FIELD_TYPE)
+INVOICE_EMAIL_COLUMN = InvoiceColumn(name=INVOICE_EMAIL_FIELD, dtype=STRING_FIELD_TYPE)
+INVOICE_ADDRESS_COLUMN = InvoiceColumn(
+    name=INVOICE_ADDRESS_FIELD, dtype=STRING_FIELD_TYPE
+)
+INSTITUTION_COLUMN = InvoiceColumn(name=INSTITUTION_FIELD, dtype=STRING_FIELD_TYPE)
+INSTITUTION_ID_COLUMN = InvoiceColumn(
+    name=INSTITUTION_ID_FIELD, dtype=STRING_FIELD_TYPE
+)
+GROUP_NAME_COLUMN = InvoiceColumn(name=GROUP_NAME_FIELD, dtype=STRING_FIELD_TYPE)
+GROUP_INSTITUTION_COLUMN = InvoiceColumn(
+    name=GROUP_INSTITUTION_FIELD, dtype=STRING_FIELD_TYPE
+)
+GROUP_BALANCE_COLUMN = InvoiceColumn(name=GROUP_BALANCE_FIELD, dtype=BALANCE_FIELD_TYPE)
+GROUP_BALANCE_USED_COLUMN = InvoiceColumn(
+    name=GROUP_BALANCE_USED_FIELD, dtype=BALANCE_FIELD_TYPE
+)
+SU_HOURS_COLUMN = InvoiceColumn(name=SU_HOURS_FIELD, dtype=INTEGER_FIELD_TYPE)
+SU_TYPE_COLUMN = InvoiceColumn(name=SU_TYPE_FIELD, dtype=STRING_FIELD_TYPE)
+SU_CHARGE_COLUMN = InvoiceColumn(name=SU_CHARGE_FIELD, dtype=BALANCE_FIELD_TYPE)
+LENOVO_CHARGE_COLUMN = InvoiceColumn(name=LENOVO_CHARGE_FIELD, dtype=BALANCE_FIELD_TYPE)
+RATE_COLUMN = InvoiceColumn(
+    name=RATE_FIELD, dtype=RATE_FIELD_TYPE
+)  # Using decimal to suppress scientific notation in export
+COST_COLUMN = InvoiceColumn(name=COST_FIELD, dtype=BALANCE_FIELD_TYPE)
+CREDIT_COLUMN = InvoiceColumn(name=CREDIT_FIELD, dtype=BALANCE_FIELD_TYPE)
+CREDIT_CODE_COLUMN = InvoiceColumn(name=CREDIT_CODE_FIELD, dtype=STRING_FIELD_TYPE)
+SUBSIDY_COLUMN = InvoiceColumn(
+    name=SUBSIDY_FIELD, dtype=BALANCE_FIELD_TYPE, default_value=Decimal(0)
+)
+BALANCE_COLUMN = InvoiceColumn(
+    name=BALANCE_FIELD,
+    dtype=BALANCE_FIELD_TYPE,
+    default_initializer=lambda df: df[COST_FIELD],
+)
+PI_BALANCE_COLUMN = InvoiceColumn(
+    name=PI_BALANCE_FIELD,
+    dtype=BALANCE_FIELD_TYPE,
+    default_initializer=lambda df: df[COST_FIELD],
+)
+
+# Internally used fields
+IS_BILLABLE_COLUMN = InvoiceColumn(name=IS_BILLABLE_FIELD, dtype=BOOL_FIELD_TYPE)
+MISSING_PI_COLUMN = InvoiceColumn(name=MISSING_PI_FIELD, dtype=BOOL_FIELD_TYPE)
+PROJECT_NAME_COLUMN = InvoiceColumn(name=PROJECT_NAME_FIELD, dtype=STRING_FIELD_TYPE)
+GROUP_MANAGED_COLUMN = InvoiceColumn(name=GROUP_MANAGED_FIELD, dtype=BOOL_FIELD_TYPE)
+CLUSTER_NAME_COLUMN = InvoiceColumn(name=CLUSTER_NAME_FIELD, dtype=STRING_FIELD_TYPE)
+IS_COURSE_COLUMN = InvoiceColumn(
+    name=IS_COURSE_FIELD, dtype=BOOL_FIELD_TYPE, default_value=False
+)
+###
+
 
 @dataclass
 class Invoice:
     export_columns_list = list()
     exported_columns_map = dict()
+    initializes_columns = tuple()
+    operates_on_columns = tuple()
 
     invoice_month: str
     data: pandas.DataFrame
@@ -77,6 +158,7 @@ class Invoice:
     export_data = None
 
     def process(self):
+        self._init_columns()
         self._prepare()
         self._process()
         self._prepare_export()
@@ -92,6 +174,24 @@ class Invoice:
     @property
     def output_s3_archive_key(self):
         return f"Invoices/{self.invoice_month}/Archive/{self.name} {self.invoice_month} {util.get_iso8601_time()}.csv"
+
+    def _init_columns(self):
+        """Initializes columns specified in `initializes_columns` and cast them to appropriate types
+
+        If column already exists, only do casting
+        If no default value is given, column initialized to None
+        """
+        for field in self.initializes_columns:
+            if field.name not in self.data.columns:
+                field_default = field.default_value
+                if field.default_initializer:
+                    field_default = field.default_initializer(self.data)
+                self.data[field.name] = field_default
+            elif self.data.dtypes[field.name] != field.dtype:
+                logger.warning(
+                    f"Column {field.name} has dtype {self.data.dtypes[field.name]} instead of expected {field.dtype}."
+                )
+            self.data = self.data.astype({field.name: field.dtype})
 
     def _prepare(self):
         """Prepares the data for processing.
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -3,7 +3,6 @@ import logging
 import os
 
 import pandas
-import pyarrow
 
 from process_report.settings import invoice_settings
 from process_report.loader import loader
@@ -22,6 +21,7 @@ from process_report.invoices import (
 )
 from process_report.processors import (
     coldfront_fetch_processor,
+    validate_input_column_processor,
     validate_pi_alias_processor,
     add_institution_processor,
     lenovo_processor,
@@ -32,6 +32,20 @@ from process_report.processors import (
     prepayment_processor,
     validate_cluster_name_processor,
 )
+
+PROCESSING_ORDER = [
+    validate_input_column_processor.ValidateInputColumnsProcessor,
+    validate_cluster_name_processor.ValidateClusterNameProcessor,
+    coldfront_fetch_processor.ColdfrontFetchProcessor,
+    validate_pi_alias_processor.ValidatePIAliasProcessor,
+    add_institution_processor.AddInstitutionProcessor,
+    lenovo_processor.LenovoProcessor,
+    validate_billable_pi_processor.ValidateBillablePIsProcessor,
+    pi_su_credit_processor.PISUCreditProcessor,
+    new_pi_credit_processor.NewPICreditProcessor,
+    bu_subsidy_processor.BUSubsidyProcessor,
+    prepayment_processor.PrepaymentProcessor,
+]
 
 
 PI_S3_FILEPATH = "PIs/PI.csv"
@@ -66,20 +80,7 @@ def main():
 
     ### Preliminary processing
     processed_data = process_merged_dataframe(
-        invoice_month,
-        merged_dataframe,
-        [
-            validate_cluster_name_processor.ValidateClusterNameProcessor,
-            coldfront_fetch_processor.ColdfrontFetchProcessor,
-            validate_pi_alias_processor.ValidatePIAliasProcessor,
-            add_institution_processor.AddInstitutionProcessor,
-            lenovo_processor.LenovoProcessor,
-            validate_billable_pi_processor.ValidateBillablePIsProcessor,
-            pi_su_credit_processor.PISUCreditProcessor,
-            new_pi_credit_processor.NewPICreditProcessor,
-            bu_subsidy_processor.BUSubsidyProcessor,
-            prepayment_processor.PrepaymentProcessor,
-        ],
+        invoice_month, merged_dataframe, PROCESSING_ORDER
     )
 
     ### Export invoices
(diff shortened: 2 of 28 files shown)


knikolla:
> @QuanMPhm Does this mean that the InitColumnsProcessor needs to know about ALL the processors?

I guess this is inevitable since we want to also upload the resulting DataFrame into Iceberg, which means we need to centralize the listing of columns, their initialization and migration.

I don't think a processor is the best way to implement this initialization. Please think about a solution that also extends the Iceberg table when a new column is introduced.

knikolla (changes requested):

naved001:
> I don't think a processor is the best way to implement this initialization. 

@knikolla Why not? I was just trying to think where else could this happen, and I guess when the CSVs are merged into the pandas dataframe before the processors run. But I want to understand your rationale behind it not being a processor.

knikolla:
> > I don't think a processor is the best way to implement this initialization.
> 
> @knikolla Why not? I was just trying to think where else could this happen, and I guess when the CSVs are merged into the pandas dataframe before the processors run. But I want to understand your rationale behind it not being a processor.

@naved001 some quick thoughts.

- The Processors are an attempt at organizing business logic into respective modules. This Processor wouldn't really contain any business logic and just perform common initialization. This strongly feels more like a base class function than a specific processor.
- Currently, each Processor already may introduce new columns, and these columns are initialized by the Processor that introduces them. This has the added benefit of providing ownership over that column and additional context and documentation about how the column is initialized and used for by the surrounding code. 
- The introduction of new columns creates ordering concerns but this ordering concern isn't fixed by the introduction of a column initializer. The business logic is what places the constraint on the ordering of Processors, not the initialization of the columns. This bug merely highlighted that as part of the PR that introduced the `PISUCreditProcessor` the initialization of that specific column should have either been moved from the `NewPICreditProcessor` to the `PISUCreditProcessor` (it now being the first one to use the column) or the ordering of the 2 processors changed (if it wouldn't make a difference to the business logic).
- We are already storing [the entire list of columns](https://github.com/CCI-MOC/invoicing/blob/main/process_report/invoices/invoice.py) in the base `Invoice` class. We can introduce more metadata to that list (for example by making each of the columns a `dataclass` containing `Column Name`, `Type`, `Default Value` or `Initializer` whatever you think is beneficial and then extend the Processors to have a `initializes_columns: List[Column]` and `operates_on_columns: List[Column]`.
- This allows us to document in code and write unit tests that check for the list of Processors and fail the unit tests when the ordering is such that a Processor `operates_on` a column that wasn't initialized by a previous Processor or doesn't have a default initializer.
- This also allows us to include within the Column definition migration logic for the Iceberg table when a column is introduced.

naved001:
@knikolla Thanks for the explanation! It makes a lot of sense to not have a processor just do the init. Just a couple of thoughts/questions.

> We are already storing [the entire list of columns](https://github.com/CCI-MOC/invoicing/blob/main/process_report/invoices/invoice.py?rgh-link-date=2026-04-03T17%3A38%3A00Z) in the base Invoice class. We can introduce more metadata to that list (for example by making each of the columns a dataclass containing Column Name, Type, Default Value or Initializer whatever you think is beneficial and then extend the Processors to have a initializes_columns: List[Column] and operates_on_columns: List[Column].

Wouldn't the processors only initialize these columns when they are called? I don't understand how this would have solved the current issue.

> This allows us to document in code and write unit tests that check for the list of Processors and fail the unit tests when the ordering is such that a Processor operates_on a column that wasn't initialized by a previous Processor or doesn't have a default initializer.

Even if we had this implemented, how would a unit test have captured the current problem which was due to the ordering of the processors? Because if we wrote the unit tests for checking if the columns exist that would be just on the fake data we are passing to that specific processor. I think the end to end tests should have been updated to capture this because we actually run all processors in a sequence and that would have failed.

knikolla:
> @knikolla Thanks for the explanation! It makes a lot of sense to not have a processor just do the init. Just a couple of thoughts/questions.
> 
> > We are already storing [the entire list of columns](https://github.com/CCI-MOC/invoicing/blob/main/process_report/invoices/invoice.py?rgh-link-date=2026-04-03T17%3A38%3A00Z) in the base Invoice class. We can introduce more metadata to that list (for example by making each of the columns a dataclass containing Column Name, Type, Default Value or Initializer whatever you think is beneficial and then extend the Processors to have a initializes_columns: List[Column] and operates_on_columns: List[Column].
> 
> Wouldn't the processors only initialize these columns when they are called? I don't understand how this would have solved the current issue.

This by itself wouldn't have solved the current issue. It would have provided a better place to centralize column data, rather than storing the column names in the base Invoice and the default initialization values in the InitProcessor.

One way it would have solved it is if a Processor `operates_on` a column that hasn't been initialized yet, the Processor would automatically set the default value at that point of the execution.

> 
> > This allows us to document in code and write unit tests that check for the list of Processors and fail the unit tests when the ordering is such that a Processor operates_on a column that wasn't initialized by a previous Processor or doesn't have a default initializer.
> 
> Even if we had this implemented, how would a unit test have captured the current problem which was due to the ordering of the processors? Because if we wrote the unit tests for checking if the columns exist that would be just on the fake data we are passing to that specific processor. I think the end to end tests should have been updated to capture this because we actually run all processors in a sequence and that would have failed.

It would have exposed the problem because the `PISUCreditProcessor` would have had the column in the `operates_on` list that it defines, however none of the preceding processors would have had the column in their `initializes_column` list. The list of the ordering of the processors is stored in code and it would be easy to have a unit test validate that this case doesn't happen.

If we decide to centralize Column information into a dataclass and have the dataclass provide a default value, we could also have the Processor at this point in the execution set that default value to the column.

naved001:
@knikolla I am convinced that having columns be dataclasses with associated metadata and defaults is a reasonable path forward. 

> The list of the ordering of the processors is stored in code and it would be easy to have a unit test validate that this case doesn't happen.

As far as I can tell, this is the only place we have a list of all the processors

https://github.com/CCI-MOC/invoicing/blob/main/process_report/process_report.py#L89-L99

I guess we just need to make sure that the tests use the same list.

knikolla:
> @knikolla I am convinced that having columns be dataclasses with associated metadata and defaults is a reasonable path forward.
> 
> > The list of the ordering of the processors is stored in code and it would be easy to have a unit test validate that this case doesn't happen.
> 
> As far as I can tell, this is the only place we have a list of all the processors
> 
> https://github.com/CCI-MOC/invoicing/blob/main/process_report/process_report.py#L89-L99
> 
> I guess we just need to make sure that the tests use the same list.

Yes, that can be stored into a global variable and the unit test can access the variable.

The end to end test would also probably throw an error since IIRC it does call all the processors.

process_report/processors/init_columns_processor.py
```diff
@@ -0,0 +1,72 @@
...
+FIELD_DEFAULT_MAPPING = {
+    invoice.BALANCE_FIELD: invoice.COST_FIELD,
+    invoice.PI_BALANCE_FIELD: invoice.COST_FIELD,
+    invoice.SUBSIDY_FIELD: Decimal(0),
+    invoice.IS_COURSE_FIELD: False,  # All rows assumed to be non-course, unless marked otherwise by ColdfrontFetchProcessor
+}
+
+
+@dataclass
+class PISUCreditProcessor(discount_processor.DiscountProcessor):
```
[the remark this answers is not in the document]
QuanMPhm: My bad

QuanMPhm:
@naved001 @knikolla I will put a hold on this and review [Jimmy's PR to add the Iceberg functionality](https://github.com/CCI-MOC/invoicing/pull/267) first, since both issues are critical and will likely influence each other. I'll see what the iceberg implementation may look like before continuing on this.

QuanMPhm:
I will continue work on this PR, now that I've done my iceberg review and draft PR(#292)

QuanMPhm:
@knikolla @naved001 I have overhauled this PR to implement @knikolla's [suggestions](https://github.com/CCI-MOC/invoicing/pull/285#issuecomment-4184431348). I encapsulated columns into a new class, and added new `Invoice` class attributes, `initializes_columns` and `operates_on_columns`. More details in the commit message.

I see that this PR is enormous, and will try to split out the type handling if the PR's too much to review before April invoicing.

I have a question below for consensus on the typing of columns.

A follow-up from this PR would be to cleanup test cases to remove redundant type casting.

process_report/invoices/invoice.py
```diff
@@ -65,18 +83,75 @@
 IS_COURSE_FIELD = "Is Course"
 ###
 
+### Initialized Column objects
+INVOICE_DATE_COLUMN = InvoiceColumn(name=INVOICE_DATE_FIELD, dtype=STRING_FIELD_TYPE)
```
QuanMPhm, line 92: @knikolla @naved001 Are you guys fine with the typing choice for these columns

naved001:
@QuanMPhm Maybe I missed some conversation, but why did you add a dedicated initColumnProcessor? This comment by @knikolla explained the reasons why we don't want that: https://github.com/CCI-MOC/invoicing/pull/285#issuecomment-4184431348

QuanMPhm:
> @QuanMPhm Maybe I missed some conversation, but why did you add a dedicated initColumnProcessor? This comment by @knikolla explained the reasons why we don't want that: [#285 (comment)](https://github.com/CCI-MOC/invoicing/pull/285#issuecomment-4184431348)

@naved001 While rewriting the PR, I decided to make a processor at the start that would check that the input invoices have all the necessary initial columns. This first processor doesn't initialize any columns by itself. It just checks that prerequisite columns exist. I guess a better name would be `ValidateInputColumnsProcessor`. I just stuck with the old name at the time of writing the PR.

QuanMPhm:
@larsks Since this PR is moreso restructuring the code rather than adding billing logic, I'd appreciate your feedback

larsks (commented):
> Since this PR is moreso restructuring the code rather than adding billing logic, I'd appreciate your feedback

@QuanMPhm There's nothing here that jumps out at me in particular. I like the standardization of using the `InvoiceColumn` dataclass to contain column metadata. It looks like you've addressed most of the comments from Kristi and Naved, and the mechanism for handling column initialization seems like an improvement.

process_report/invoices/invoice.py
naved001: It's fine by me.

process_report/invoices/invoice.py
```diff
@@ -1,9 +1,27 @@
...
+@dataclass
+class InvoiceColumn:
+    name: str
+    dtype: Any
+    default_value: Any | None = None
+    default_initializer: Callable[[pandas.DataFrame], pandas.Series] | None = None
+
+
+# Field type definitions
+BALANCE_FIELD_TYPE = pandas.ArrowDtype(pyarrow.decimal128(21, 2))
```
naved001, line 23: Optional: I'd call this Decimal Field Type (the others are also named based on the type, this is only one that's using the purpose in the name).

process_report/processors/validate_input_column_processor.py
```diff
@@ -0,0 +1,37 @@
...
+            for column in self.initializes_columns
+            if column.name not in self.data.columns
+        ]
+        if missing_columns:
+            raise ValueError(
+                f"Input dataframe is missing required columns: {', '.join(missing_columns)}. Stopping invoicing"
+            )
+
+        # Casts columns to appropriate types
+        self._init_columns()
```
naved001, line 37: It appears that this processor doesn't initialize the processors but it does cast the the columns defined. My question is where does this casting currently happen?

process_report/processors/init_columns_processor.py
```diff
@@ -0,0 +1,37 @@
+from dataclasses import dataclass
+
+from process_report.invoices import invoice
+from process_report.processors import processor
+
+
+@dataclass
+class InitColumnsProcessor(processor.Processor):
```
naved001: Don't you think the name is misleading? It doesn't actually initialize the columns it only casts them. This is a cause of confusion imo

process_report/tests/unit/processors/test_processor_list.py
```diff
@@ -0,0 +1,26 @@
+from process_report.process_report import PROCESSING_ORDER
+from process_report.tests.base import BaseTesCase
```
naved001: I guess this typo "BaseTesCase" is everywhere because your IDE saved the day. Could be solved in a different PR if it wasn't introduced in this one.

naved001 (commented):
Mostly looks fine to me but a few questions in line. I need to take a deep look at the tests as well. Thanks!

process_report/invoices/invoice.py
QuanMPhm: I wanted to differentiate between the balance and rate field. The `Rate` column, which contains the SU rates, like `0.0000087890625`, is `Decimal`. Since `BALANCE_FIELD_TYPE` is specifically for monetary columns, I figured `BALANCE_FIELD_TYPE` was the best name.

And I just realized this PR is not updated, so my comment above might not make much sense. I'll push up the latest updates I made and you'll see what I mean

process_report/processors/validate_input_column_processor.py
QuanMPhm: The current code [only casts two columns](https://github.com/CCI-MOC/invoicing/blob/72b6e76e06c55e28a96af46bc61b8f2d65db6892/process_report/process_report.py#L111-L114) in `merge_csv()`. It currently does not systematically cast all columns, or checks if they exist.

process_report/processors/init_columns_processor.py
QuanMPhm: renamed to `ValidateInputColumnsProcessor`

process_report/processors/prepayment_processor.py
```diff
@@ -20,14 +20,31 @@ class PrepaymentProcessor(discount_processor.DiscountProcessor):
...
+        invoice.GROUP_MANAGED_COLUMN,
+        invoice.GROUP_BALANCE_COLUMN,
+        invoice.GROUP_BALANCE_USED_COLUMN,
+    )
+    operates_on_columns = (
+        invoice.GROUP_NAME_COLUMN,
+        invoice.GROUP_INSTITUTION_COLUMN,
+        invoice.GROUP_MANAGED_COLUMN,
+        invoice.GROUP_BALANCE_COLUMN,
+        invoice.GROUP_BALANCE_USED_COLUMN,
```
naved001: I realized that a processors will always operate on columns that it initalizes i.e. `initializes_columns` is a subset of `operates_on_columns`. You could do this to cut down on the repetition. Just a suggestion.
```python
operates_on_columns = (
    *initalizes_columns,
    invoice.INVOICE_EMAIL_COLUMN,
	...
	...
)
```

process_report/tests/e2e/test_data/test_pi.yaml
```diff
@@ -1,4 +1,4 @@
 - username: PI9
-- username: PI10
+- username: pi2@harvard.edu
   non_billed_su_types:
-    - name: SU1
+    - name: Free CPU
```
naved001, line 4: why did you make this change in this PR?

process_report/invoices/invoice.py
```diff
@@ -93,6 +171,20 @@ def output_s3_key(self) -> str:
     def output_s3_archive_key(self):
         return f"Invoices/{self.invoice_month}/Archive/{self.name} {self.invoice_month} {util.get_iso8601_time()}.csv"
 
+    def _init_columns(self):
+        """Initializes columns specified in `initializes_columns` and cast them to appropriate types
+
+        If column already exists, only do casting
```
knikolla, line 181: I would (at a minimum as it arguably should be an error) log a warning here for this case specifically: in case the column already exists and is of a different type. The column should have been created of the right type from the get go. It having been created of the wrong type and requiring casting might silently hide a potential loss of accuracy or data.

knikolla (changes requested):
Great work! Only one comment otherwise direction and everything looks good.

process_report/invoices/invoice.py
QuanMPhm: Adding the warning as you suggested, I think the output is not very helpful:
```
WARNING:process_report.invoices.invoice:Column Invoice Month already exists in data but has dtype str instead of expected string.
WARNING:process_report.invoices.invoice:Column Project - Allocation already exists in data but has dtype str instead of expected string.
WARNING:process_report.invoices.invoice:Column Project - Allocation ID already exists in data but has dtype str instead of expected string.
WARNING:process_report.invoices.invoice:Column Manager (PI) already exists in data but has dtype object instead of expected string.
WARNING:process_report.invoices.invoice:Column Cluster Name already exists in data but has dtype str instead of expected string.
WARNING:process_report.invoices.invoice:Column Invoice Email already exists in data but has dtype float64 instead of expected string.
WARNING:process_report.invoices.invoice:Column Invoice Address already exists in data but has dtype float64 instead of expected string.
WARNING:process_report.invoices.invoice:Column Institution already exists in data but has dtype float64 instead of expected string.
WARNING:process_report.invoices.invoice:Column Institution - Specific Code already exists in data but has dtype object instead of expected string.
WARNING:process_report.invoices.invoice:Column SU Hours (GBhr or SUhr) already exists in data but has dtype int64 instead of expected int64[pyarrow].
WARNING:process_report.invoices.invoice:Column SU Type already exists in data but has dtype str instead of expected string.
WARNING:process_report.invoices.invoice:Column Rate already exists in data but has dtype float64 instead of expected decimal128(21, 13)[pyarrow].
```

At the start, I convert the columns to pandas dtypes. Alot of them get converted from the default python types to their pandas equivalent (`str` to pandas `string`, `int` to `int64[pyarrow]`). I believe the cases where data loss may happen would be the casting to `Decimal`, in which case I think pandas does a pretty good job at casting and rounding:

```
TEST_INVOICE = """
Cost
0.115
0.114
1.11
"""

a = pandas.read_csv(io.StringIO(TEST_INVOICE))
a.loc[2, "Cost"] = a.loc[2, "Cost"] + 2.221
print(a.to_csv(index=False))

# Cost
# 0.115
# 0.114
# 3.3310000000000004

a = a.astype({"Cost": pandas.ArrowDtype(pyarrow.decimal128(21, 2))})
print(a.to_csv(index=False))

# Cost
# 0.12
# 0.11
# 3.33
```

Do you still want me to include the warning log?

knikolla: > Alot of them get converted from the default python types to their pandas equivalent (str to pandas string, int to int64[pyarrow])

When you load the CSV files, you should explicitly provide the column data types.

> I believe the cases where data loss may happen would be the casting to Decimal, in which case I think pandas does a pretty good job at casting and rounding:

Why risk it on a "pretty good job" when you can explicitly load the CSV file with the correct typing of the column from the get go? Especially now that we have defined each type of each column centrally.

knikolla: I think limiting the warning to conversions to Decimal, from something else is enough for now.

I'm confused by
```
WARNING:process_report.invoices.invoice:Column Rate already exists in data but has dtype float64 instead of expected decimal128(21, 13)[pyarrow].
```

because we load the column Rate as a string. https://github.com/CCI-MOC/invoicing/blob/72b6e76e06c55e28a96af46bc61b8f2d65db6892/process_report/process_report.py#L107-L117

Now that we have a defined Rate data type we should probably take advantage of that.

QuanMPhm: Sorry, when generating those warning logs, I was also doing additionally testing, specifically with the code you quoted. I removed the casting so the code looked like this:
```
dataframe = pandas.read_csv(
    file,
    engine="pyarrow",
    quotechar="|",
)
```

I'll implement your feedback now

process_report/tests/e2e/test_data/test_pi.yaml
QuanMPhm: This is to allow the e2e test to run more of the billing code, specifically code in `PISUCreditProcessor` that caused the billing bug we had a while back. In this case, the test data has been changed to give a PI the PI-SU credit

process_report/processors/prepayment_processor.py
QuanMPhm: Didn't even realize I could do that

larsks: That is called "list unpacking". You often see this when passing a list of arguments to a function that expects multiple arguments. E.g., if I have:

```python
def example(greeting: str, noun: str):
  print(f"{greeting.title()}, {noun}!")
```

Then I can do this:

```python
>>> example("hello", "world")
Hello, world!
```

But if I had the arguments in a list, I could do this:

```python
>>> args=["hello", "world"]
>>> example(*args)
```

Using the `*` operator like that isn't unique to argument lists, which is why you can do things like @naved001 suggested:

```python
>>> list1 = ["cyan", "magenta"]
>>> list2 = [*list1, "yellow", "black"]
>>> list2
['cyan', 'magenta', 'yellow', 'black']
```

Although you can also accomplish the same thing with list addition:

```python
>>> list1 = ["cyan", "magenta"]
>>> list2 = list1 + ["yellow", "black"]
>>> list2
['cyan', 'magenta', 'yellow', 'black']
```

There is also dictionary unpacking through the `**` operator. You've probably
seen this in the context of passing keyword arguments to a function from a
dictionary. Given:

```python
def example(name:str|None = None, company:str|None = None):
  print(f"{name.title()} works at {company.title()}")
```

We can call it like this:

```python
>>> args={'name': 'Lars', 'company': 'red hat'}
>>> example(**args)
Lars works at Red Hat
```

This can also be used for extending dictionaries:

```python
>>> dict1 = {'name': 'lars', 'company': 'redhat'}
>>> dict2 = {**dict1, 'cats': 3}
>>> dict2
{'name': 'lars', 'company': 'redhat', 'cats': 3}
```

Or even merging dictionaries:

```python
>>> dict1 = {'name': 'lars', 'company': 'redhat'}
>>> dict2 = {'town': 'belmont', 'state': 'ma'}
>>> dict3 = {**dict1, **dict2}
>>> dict3
{'name': 'lars', 'company': 'redhat', 'town': 'belmont', 'state': 'ma'}
```

Although since Python 3.9, the more obvious way of doing this would be using the `|` union operator:

```python
>>> dict1 = {'name': 'lars', 'company': 'redhat'}
>>> dict2 = {'town': 'belmont', 'state': 'ma'}
>>> dict3 = dict1 | dict2
```

QuanMPhm: Thank you for the very thorough tutorial! I saw it used in function calls, not in a list definition before.

process_report/invoices/invoice.py
QuanMPhm: I decided to let the warning be fired on any type conversion. Now that we're specifying all dtypes upon loading the CSV, the warning should never be seen in normal execution.

QuanMPhm:
@knikolla @naved001 I have responded to all comments.

process_report/process_report.py
```diff
@@ -109,8 +110,19 @@ def merge_csv(files):
             file,
             engine="pyarrow",
             dtype={
-                invoice.COST_FIELD: pandas.ArrowDtype(pyarrow.decimal128(21, 2)),
-                invoice.RATE_FIELD: str,
+                invoice.INVOICE_DATE_FIELD: invoice.STRING_FIELD_TYPE,
```
knikolla: You have already defined the columns into a `InvoiceColumn` class containing both `name` and `dtype`. Therefore I strongly suggest you use that pairing here too, otherwise you are maintaining 2 copies of what type a column is and those are prone to go out of sync with time. 

such as 
```python
dtype={
    invoice.RATE_COLUMN.name: invoice.RATE_COLUMN.dtype,
...
}
```

knikolla (changes requested):

knikolla (approved):
Given that a warning will be output if a column's dtype is different from the type defined in the `InvoiceColumn` type, I'm fine with my requested change in the previous review being done in a follow-up PR.

process_report/processors/validate_billable_pi_processor.py
```diff
@@ -95,6 +95,16 @@ class ValidateBillablePIsProcessor(processor.Processor):
...
 
+    initializes_columns = [invoice.IS_BILLABLE_COLUMN, invoice.MISSING_PI_COLUMN]
+    operates_on_columns = [
+        *initializes_columns,
+        invoice.PI_COLUMN,
+        invoice.PROJECT_COLUMN,
+        invoice.CLUSTER_NAME_COLUMN,
+        invoice.IS_COURSE_COLUMN,
+        invoice.INSTITUTION_COLUMN,
+    ]
```
naved001: these are lists here and not tuples like everywhere else.

naved001 (approved):
Just one more thing, but I am approving the PR so merge it once that's addressed.

knikolla:
@QuanMPhm before merging, please create issues to track the 2 outstanding requested changes.

QuanMPhm:
@knikolla @naved001 I've implemented both of your feedback to spare everyone's time

PR #290, opened 2026-04-13
Split up test_discount_processor.py into tests for the discount processor and pi-su credit processor

marcoagonzales007:
-Moved `test_one_eligible_project_only` and `test_all_eligible_projects` into a new `test_pi_su_credit_processor.py` because they test the `PISUCreditProcessor._process` logic 

-Kept `test_preexisting_credit` in `test_discount_processor.py` 

-Added `new_pi_su_credit_processor()` to `tests/util.py` to match the other processors

-Updated both of the test files to use `test_utils` instead of directly using `PISUCreditProcessor`

Closes Issue #289

### DIFF
--- a/process_report/tests/unit/processors/test_discount_processor.py
+++ b/process_report/tests/unit/processors/test_discount_processor.py
@@ -3,7 +3,7 @@ from unittest import TestCase
 import pandas
 
 from process_report.invoices import invoice
-from process_report.processors.pi_su_credit_processor import PISUCreditProcessor
+from process_report.processors.discount_processor import DiscountProcessor
 
 
 class TestDiscountProcessor(TestCase):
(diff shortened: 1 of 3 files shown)


process_report/tests/unit/processors/test_discount_processor.py
```diff
@@ -52,7 +52,7 @@ def test_preexisting_credit(
             balance=[90],
         )
 
-        processor = PISUCreditProcessor(
+        processor = test_utils.new_pi_su_credit_processor(
```
QuanMPhm: Since this file only tests the features of `DiscountProcessor`. You should not be using the New-PI processor here. I would suggest using the `DiscountProcessor` and call `apply_flat_discount()` directly instead. Something like this:

```python
processor = DiscountProcessor(data=pandas.Dataframe())
processor.apply_flat_discount(...)
...
```

You shouldn't need to add an function for `DiscountProcessor` in `util.py`, since its only used once here.

QuanMPhm (changes requested):
Aside from the comment below, can you add your PR description into the commit message as well?

.gitignore
```diff
@@ -3,3 +3,4 @@
 !process_report/tests/e2e/test_data/test_invoices/*.csv
 __pycache__/
 *.py[cod]
+.vscode/
```
QuanMPhm: I assume this is something Visual Studio Code added to the repo when you worked on it, in which case I would suggest not putting this here, since it doesn't benefit anyone else developing on this repository. If everyone added things to `.gitignore` that's specific to their work setup, the file will become messy quick.

If you want a personal ignore rules for your fork, I'd suggest [something described here](https://dev.to/fronkan/a-personal-gitignore-even-for-a-single-repository-4o7h)

QuanMPhm (commented):
One minor comment. Otherwise, looks good!

QuanMPhm (approved):

QuanMPhm:
@knikolla @naved001 Looks good to me. Just a pass from either of you

PR #291, opened 2026-04-17
Move nonbillable pydantic models from invoicing-private-data

### LINKED ISSUE #142: Move validation code from `invoicing-private-data` to here
## Motivation
As discussed in https://github.com/CCI-MOC/invoicing-private-data/issues/92, `invoicing-private-data` now has a few highly structured YAML files with pydantic models to validate them. We should move those models to this repo to avoid duplicating validation logic, which is the case now in `loader.py`, such as [with loading the nonbillable project list](https://github.com/CCI-MOC/invoicing/blob/72b6e76e06c55e28a96af46bc61b8f2d65db6892/process_report/loader.py#L139-L148). This will also enforce separation between private business data and open source code.

This task is low priority, and I'll likely assign it to the new intern.

CC @knikolla @naved001 

## Completion Criteria
The pydantic models are defined and used in this repo. `invoicing-private-data` should then install `invoicing` to validate its YAML files

## Description
- [ ] PR in `invoicing` merged
- [ ] PR in `invoicing-private-data` merged (Depends on first task)

## Completion dates
Desired - TBD
Required - TBD

### LINKED ISSUE #287: Move validation code from `invoicing-private-data` to here
## Motivation
As discussed in https://github.com/CCI-MOC/invoicing-private-data/issues/92, `invoicing-private-data` now has a few highly structured YAML files with pydantic models to validate them. We should move those models to this repo to avoid duplicating validation logic, which is the case now in `loader.py`, such as [with loading the nonbillable project list](https://github.com/CCI-MOC/invoicing/blob/72b6e76e06c55e28a96af46bc61b8f2d65db6892/process_report/loader.py#L139-L148). This will also enforce separation between private business data and open source code.

This task is low priority, and I'll likely assign it to the new intern.

CC @knikolla @naved001 

## Completion Criteria
The pydantic models are defined and used in this repo. `invoicing-private-data` should then install `invoicing` to validate its YAML files

## Description
- [ ] PR in `invoicing` merged
- [ ] PR in `invoicing-private-data` merged (Depends on first task)

## Completion dates
Desired - TBD
Required - TBD

marcoagonzales007:
-Copied the pydantic models from `invoicing-private-data/validate_projects/models.py` into a new `process_report/nonbillable_models.py` in this repo.
-Updated loader.py to validate the YAML through the pydantic models instead of using raw dicts 
Closes CCI-MOC/MOC-issues#142


process_report/loader.py
```diff
@@ -147,44 +146,43 @@ def get_nonbillable_projects(self) -> pandas.DataFrame:
...
-                if not _is_in_time_range(project):
+            data = yaml.safe_load(file)
+        projects = ExcludedProjectList.model_validate(data)
+        for project in projects.root:
+            project_name = project.name
+            cluster_list = project.clusters.root
+            is_billable = project.is_billable
+
+            if project.start:
+                if not _is_in_time_range(project.start, project.end):
```
QuanMPhm: Could you move all the parsing logic here into the `ExcludedProjectList` models? Same goes for the PI-list parsing logic as well. They should have a function such that given the invoice month, returns both the list of timed, and non-timed projects. It makes the most sense for the models to have the parsing/validation logic, and for `loader.py` to only handle "loading".

Afterwards, can you move the test cases for the models over here as well? Place them in a unit test file

pyproject.toml
QuanMPhm, line 1: Making `invoicing` a installable package is an independant change from moving the models. I would suggest moving this file into a seperate PR instead. That way, in case this PR gets delayed (for any reason, hypothetically), we can still move forward with adding the `pyproject.yaml`

QuanMPhm (changes requested):
A few structural comments

process_report/models/cluster_names.txt
```diff
@@ -0,0 +1,8 @@
+Cluster1
+Cluster2
```
QuanMPhm: I see that these names need to be added to pass the unit tests, but these aren't actually cluster names we're using in production. You don't want to clutter the real data with fake test data

Creating fake data for the purpose of testing is a common programming task, so there are many of existing tools to help you create fake data without effecting the "real" data. For `invoicing` and many of oru other repos, we use [`unittest.mock`](https://docs.python.org/3/library/unittest.mock.html), or one of `pytest`'s numerous mocking or fixture tools, like [`parametrize`](https://docs.pytest.org/en/stable/how-to/parametrize.html). 

I'll give you more detailed suggestions later down this review on how you should mock the data in the case of this PR

process_report/models/nonbillable_models.py
```diff
@@ -0,0 +1,185 @@
...
+    non_billed_su_types: NonBilledSUTypeList | None = None
+
+    model_config = pydantic.ConfigDict(populate_by_name=True)
+
+
+PIList = UniqueObjectList[PIParticipant]
+
+
+def get_nonbillable_pis(pi_list: PIList) -> list[str]:
+    return [pi.name for pi in pi_list.root if pi.non_billed_su_types is None]
```
QuanMPhm: This function, as well as `get_pi_non_billed_su_types()`, `get_nonbillable_projects()`, and `get_nonbillable_timed_projects` in `util.py`, should be moved inside the Pydantic model itself, instead of being top-level functions. Doing so documents in code that these are the functionalities of the models themselves. Being instance methods may also simplify them a bit. See how it goes rewriting them as model methods.

process_report/models/su_types.txt
```diff
@@ -0,0 +1,23 @@
+SU1
```
QuanMPhm: See earlier comment on mocking fake data

process_report/tests/unit/test_nonbillable_models.py
QuanMPhm, line 1: Follow-up on the comment about some functions that should be moved in the pydantic models themselves, this file should now contain test cases for the nonbillable list that has been sitting in [`test_util.py`](https://github.com/CCI-MOC/invoicing/blob/72b6e76e06c55e28a96af46bc61b8f2d65db6892/process_report/tests/unit/test_util.py#L108-L140). 

When you move the test cases over here, you should be able to use `pytest.parametrize` to help you create  test data for the nonbillable list, and `pytest-mock` to mock `get_allowed_clusters()`

process_report/nonbillable_models.py
QuanMPhm, line 1: This file is a duplicate of `models/nonbillable_models.py`. Is there a reason you kept a copy here?

pyproject.toml
QuanMPhm: This file is still here

QuanMPhm (changes requested):
Looking better. A few comments

process_report/nonbillable_models.py
marcoagonzales007: Sorry I must have  messed up the files in my commit and thats why the toml file is still there as well.

process_report/tests/unit/test_nonbillable_models.py
```diff
@@ -0,0 +1,230 @@
...
+            },
+            {"name": "ProjectE"},
+        ],
+    ],
+)
+def test_timed_projects(mocker, projects_data):
+    mocker.patch(
+        "process_report.models.nonbillable_models.get_allowed_clusters",
+        return_value={"Cluster1", "Cluster2"},
+    )
```
QuanMPhm: The test project list here, as well as the `mock.patch` call, is repeated in the function below. If there's set-up that multiple test functions use, you can use `pytest` fixtures, even for mocking. An example would be:

```python suggestion
@pytest.fixture
def sample_project_list():
    return [
        {
            "name": "ProjectA",
            "clusters": [{"name": "Cluster1"}, {"name": "Cluster2"}],
        },
        {
            "name": "ProjectB",
            "clusters": [
                {"name": "Cluster1", "start": "2023-01", "end": "2023-12"}
            ],
        },
        {"name": "ProjectC", "start": "2023-06", "end": "2023-07"},
        {
            "name": "ProjectD",
            "is_billable": True,
            "clusters": [
                {"name": "Cluster1", "start": "2023-05", "end": "2023-09"},
                {"name": "Cluster2", "start": "2023-05", "end": "2023-11"},
            ],
        },
        {"name": "ProjectE"},
    ]

@pytest.fixture
def mock_allowed_clusters(mocker):
    return mocker.patch(
        "process_report.models.nonbillable_models.get_allowed_clusters",
        return_value={"Cluster1", "Cluster2"},
    )

def test_timed_projects(mock_allowed_clusters, sample_project_list):
    projects = nonbillable_models.ExcludedProjectList.model_validate(sample_project_list)
    ...
```

process_report/loader.py
```diff
@@ -113,27 +114,20 @@ def load_dataframe(self, filepath: str) -> pandas.DataFrame:
         return pandas.read_csv(filepath)
 
     @functools.lru_cache
-    def _load_pi_config(self, filepath: str) -> list[dict]:
+    def _load_pi_config(self, filepath: str) -> PIList:
         with open(filepath) as file:
             pi_list = yaml.safe_load(file)
```
QuanMPhm: Since `_load_pi_config` will only ever receive `invoice_settings.nonbillable_pis_filepath` as the file to load, could you do this instead?

```python suggestion
def _load_pi_config(self) -> PIList:
        with open(filepath) as file:
            pi_list = yaml.safe_load(invoice_settings.nonbillable_pis_filepath)

        return PIList.model_validate(pi_list)

    def get_nonbillable_pis(self) -> list[str]:
        return self._load_pi_config().get_nonbillable_pis()

    def get_pi_non_billed_su_types(self) -> dict[str, list[str]]:
        """PI usernames -> list of SU types that receive credit (zeroed out)."""
        eturn self._load_pi_config().get_pi_non_billed_su_types()
```

QuanMPhm (changes requested):
Just two more comments

QuanMPhm (changes requested):
The code changes looks good. Can you squash all your commits into one, and write a concise commit message?

requirements.txt
```diff
@@ -6,3 +6,4 @@ Jinja2
 validators
 python-dateutil
 pydantic-settings
+pytest-mock
```
QuanMPhm: I just realized, can you put this into the [`test-requirement.txt`](https://github.com/CCI-MOC/invoicing/blob/main/process_report/tests/test-requirements.txt) file, as this is a dependancy for tests?

QuanMPhm (approved):
One minor comment. Otherwise, this looks good. @knikolla @naved001 I believe this is good enough for your review

QuanMPhm (changes requested):
It seems the recently merged PR introduced some merge conflicts. Please resolve them

QuanMPhm (approved):
Looks good!

PR #292, opened 2026-04-20
Enable exporting monthly invoice to Iceberg

### LINKED ISSUE #144: Feature: Append dataframe to Iceberg during invoicing pipeline run
With the ongoing Iceberg implementation CCI-MOC/invoicing#252 CCI-MOC/MOC-issues#145 , we will need a function that exports the finalized dataframe into parquet files so they can be appended to the iceberg table.

I am wondering however upon normal circumstances going forward i.e. (at the stage where the tables have already been backfilled) would it be useful to have a flag or a function that stops the generation of the downstream files - i.e., `.csv`, `.pdf` etc? Because during the backfilling process I will not be doing all the downstream file creation because that  process is quite slow, especially pdf generation.

@knikolla @QuanMPhm

### LINKED ISSUE #259: Feature: Export dataframe to parquet and upload to Iceberg during invoicing pipeline run
With the ongoing Iceberg implementation #252 #258 , we will need a function that exports the finalized dataframe into parquet files so they can be appended to the iceberg table.

I am wondering however upon normal circumstances going forward i.e. (at the stage where the tables have already been backfilled) would it be useful to have a flag or a function that stops the generation of the downstream files - i.e., `.csv`, `.pdf` etc? Because during the backfilling process I will not be doing all the downstream file creation because that  process is quite slow, especially pdf generation.

@knikolla @QuanMPhm

QuanMPhm:
Closes CCI-MOC/MOC-issues#144. This PR is in draft, since I'm not too clear which dataframe we want to export. Also, I want consensus on the datatypes of the columns we want in the iceberg table. It seems this PR will depend on CCI-MOC/invoicing#285. 

@knikolla @naved001 Regardless, for now, this draft should be enough to show you guys what the iceberg operations involved will look like, and the structure of code changes that will need to happen

### DIFF
--- a/process_report/invoices/iceberg_invoice.py
+++ b/process_report/invoices/iceberg_invoice.py
@@ -1,0 +1,82 @@
+import logging
+from dataclasses import dataclass, field
+
+from pyiceberg.table import Table
+from pyiceberg.catalog import Catalog, load_catalog
+import pyarrow
+
+import process_report.invoices.invoice as invoice
+from process_report.loader import loader
+from process_report.settings import invoice_settings
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+
+def get_iceberg_catalog(config: dict, catalog_name: str) -> Catalog:
+    return load_catalog(name=catalog_name, **config)
+
+
+def get_iceberg_table(catalog: Catalog, table_path) -> Table:
+    return catalog.load_table(table_path)
+
+
+@dataclass
+class IcebergInvoice(invoice.Invoice):
+    export_columns_list = [
+        invoice.INVOICE_DATE_FIELD,
+        invoice.PROJECT_FIELD,
+        invoice.PROJECT_ID_FIELD,
+        invoice.PI_FIELD,
+        invoice.CLUSTER_NAME_FIELD,
+        invoice.INVOICE_EMAIL_FIELD,
+        invoice.INVOICE_ADDRESS_FIELD,
+        invoice.INSTITUTION_FIELD,
+        invoice.INSTITUTION_ID_FIELD,
+        invoice.IS_BILLABLE_FIELD,
+        invoice.SU_HOURS_FIELD,
+        invoice.SU_TYPE_FIELD,
+        invoice.RATE_FIELD,
+        invoice.GROUP_NAME_FIELD,
+        invoice.GROUP_INSTITUTION_FIELD,
+        invoice.GROUP_BALANCE_FIELD,
+        invoice.COST_FIELD,
+        invoice.GROUP_BALANCE_USED_FIELD,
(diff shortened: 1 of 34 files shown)


jimmysway:
Aren't the dataframes that we'll be exporting just the processed billable and non-billable dataframes?

jimmysway:
Is Billable, Cluster Assignment Source, Cluster Assignment Method

Are three new columns that have been created from the backfilling process. going forward. We can just set Cluster Assignment Source and Cluster Assignment method to Something like "Regular Monthly Invoicing Cycle" or something like that.

There is already an `Is Billable` column or flag or indicator somewhere in the invoicing code it is just a matter of exporting that as well instead of separating billable from non-billable. I think the plan was to ideally export the entire combined dataframe (billable + nonbillable) to the iceberg.

@QuanMPhm @knikolla

QuanMPhm:
> We can just set Cluster Assignment Source and Cluster Assignment method to Something like "Regular Monthly Invoicing Cycle" or something like that.

I am hesistant to include those columns in the Iceberg table. I believe they should just be removed, since they're not useful for anything in invoicing. We had those columns in the first place because it helped in constructing the initial iceberg table. 

As for what invoice to export, I fine with exporting the entire "main" dataframe, since that's equivalent to billable + nonbillable, and it is also what's currently in the initial iceberg table. 

@knikolla @naved001 I'll wait for your review

knikolla:
We shouldn't have `Cluster Assignment Source`, `Cluster Assignment Method` in the Iceberg table

`Is Billable` should be there.

QuanMPhm:
@knikolla @naved001 @jimmysway I now consider this PR ready for review. I have rebased it on top of the [column tracking PR](https://github.com/CCI-MOC/invoicing/pull/285), as the current state of the e2e test won't allow the invoice to be exported to iceberg easily. Specifically, Iceberg does not allow exporting of columns with a `pa.null` type, which happens when a column only contains `None`. The current e2e test does cause some columns to only be `None`. Since the column tracking PR ensures every column has a non-null type, the issue is removed.

QuanMPhm:
@knikolla @naved001 @jimmysway I just learned that `pyiceberg` [does not support interfacing with hadoop catalogs](https://github.com/apache/iceberg-python/issues/17), which is what our current iceberg table in s3 is using. So I'll need to rewrite my PR to use something like `PySpark`.

jimmysway:
> @knikolla @naved001 @jimmysway I just learned that `pyiceberg` [does not support interfacing with hadoop catalogs](https://github.com/apache/iceberg-python/issues/17), which is what our current iceberg table in s3 is using. So I'll need to rewrite my PR to use something like `PySpark`.

What's the status on this? I am happy to take this over because pyspark is what I used to do the backfill

PR #294, opened 2026-05-14
Updated Dockerfile and manifests for invoice cronjob

### LINKED ISSUE #1514: Capture non-billable projects as they are created so we no longer have to review them as part of invoicing automation
## Motivation
Invoicing automation is supposed to be automated and therefore quick. If we can front-load confirming non-billable when the projects are created, we will allow the automation to be automated.

## Completion Criteria
Project creation includes confirming if it is non-billable.

## Description
- [x] Review if this can be done in ColdFront, if not, where does this need to be done?
- [ ] ensure wherever we need to update gets updated.

## Completion dates
Desired - 2026-04-20
Required - TBD

QuanMPhm:
Closes https://github.com/nerc-project/operations/issues/1514. Dockerfile, entrypoint script and cronjob has been updated. A new daily cronjob has been added to check if invoicing pipeline runs without error. The daily cronjob will not update to S3.

This is a step towards automating the invoicing pipeline and reducing manual work.

@knikolla @naved001 Do you want a new Github action workflow that runs the invoicing pipeline daily to check, or how else do you want to be notified by the invoicing pipeline encounters an error?


k8s/base/invoice-daily-check.yaml
```diff
@@ -0,0 +1,42 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: invoice-daily-check
+spec:
+  schedule: "0 14 * * *"
```
knikolla: Tasks like this, generally are referred to as "nightly" and should typically be run outside of times when there are users and load on the systems. Running this at 9 or 10am as you are doing is generally not advised and something like 3am would be preferable. 

At that time nobody would be interacting with Keycloak and ColdFront, which this would be generating additional load to.

knikolla (changes requested):
Looks good, one comment.

Also, which OpenShift cluster are we are going to be running this on?

k8s/base/invoice-daily-check.yaml
QuanMPhm: Updated run time to 3AM Boston time. Initially, I chose the time based on what was in [the openshift usage script repo](https://github.com/CCI-MOC/openshift-usage-scripts/blob/main/k8s/base/produce-report-cronjob.yaml), I guess you changed the runtime for that cronjob manually?

naved001: @QuanMPhm oh right, what's in the repo is kind of an example. The openshift reports show up around 2AM.

tools/setup_and_process.sh
```diff
@@ -27,17 +27,21 @@ if [ ! -d ./non-billable-projects ]; then
...
-    --nonbillable-file "NERC (Nonbillable)" \
-    --output-file "NERC" \
-    --output-folder "PI Invoices" \
-    --BU-invoice-file "NERC BU" \
-    --Lenovo-file "Lenovo" \
-    --BU-subsidy-amount 100
+
+export S3_ENDPOINT='https://s3.us-east-005.backblazeb2.com'
+export S3_APP_KEY
+export S3_KEY_ID
```
larsks: What are these empty `export` statements for? These variables don't seem to be set elsewhere. It's not clear why we're exporting unset variables.

Also, should some of these be optionally inherited from the external environment? E.g.:

```
export S3_ENDPOINT=${S3_ENDPOINT:-'https://s3.us-east-005.backblazeb2.com'}
```

That way the value could be set in the Dockerfile or on the `docker run` command line (`docker run -e S3_ENDPOINT=...`).

QuanMPhm: `S3_APP_KEY` and  `S3_KEY_ID` are set in the cronjob manifests, such as in `k8s/base/invoice-processing.yaml`. You made me go and test my manifests, and realized I made some mistakes. The manifests have been amended

larsks: Okay, but my question remains: what are these empty `export` statements for? The variables are already in the environment.

QuanMPhm: I did some more testing and I now see that you're right. Initially, I assumed that the environment variables had to be exported for the python script to access them, but apparently not in the case of a Docker container? Or Pod? I tested using a pod on my local Openshift.

I removed all the empty export statements

larsks: > Initially, I assumed that the environment variables had to be exported for the python script to access them

The shell `export` command takes a shell variable and exports it to the environment, where it will be available to subprocesses. A shell variable is something you create with an assignment statement in your shell script:

```sh
myvar="hello world"
```

That variable is local to the shell; it won't be visible in subprocesses:

```sh
$ myvar="hello world"
$ python -c 'import os; print(os.environ.get("myvar", "<does not exist>"))'
<does not exist>
```

The `export` command promotes the shell variable into the environment, where it will be inherited by subprocesses:

```sh
$ myvar="hello world"
$ export myvar
$ python -c 'import os; print(os.environ.get("myvar", "<does not exist>"))'
hello world
```

Your variables are *already* set in the inherited environment. By setting values in the `env` section of your CronJob manifest, you are setting up the inherited environment of the entire container; *every* process spawned inside that container will already have those variables in its environment. You have nothing to `export` because you have not created any shell-local variables.

This is effectively the same behavior you get when running a container with the `-e` command line option (or `--env-file`, if you prefer):

```sh
$ podman run -e myvar="hello world" docker.io/python:3.14 python -c 'import os; print(os.environ.get("myvar", "<does not exist>"))'
hello world
```

naved001: @larsks 

> $ myvar="hello world"
> $ export myvar

I never realized I was doing 2 things because I would always do `export myvar="hello world"` . Thanks for the write up!

k8s/base/invoice-daily-check.yaml
```diff
@@ -0,0 +1,42 @@
...
+        spec:
+          containers:
+          - name: invoice-daily-check
+            image: ghcr.io/cci-moc/process-csv-report:latest
+            env:
+              - name: GH_NONBILLABLE_DEPLOYKEY
+                valueFrom:
+                secretKeyRef:
+                  name: gh-nonbillables
+                  key: ssh-deploykey
```
naved001: the key in the secret for the other job is called ssh-privatekey ( and you changed the sample secret file to match that). There's no ssh-deploykey anywhere.

naved001 (commented):
one comment but otherwise looks fine to me. I can deploythis on the nerc-ocp-prod cluster (where I run all the openshift jobs)

naved001 (approved):

knikolla (approved):

PR #295, opened 2026-05-18
Add unit tests for AddInstitutionProcessor to prepare invoicing for complete automation

marcoagonzales007:
part of CCI-MOC/MOC-issues#147 

Adds unit tests for `AddInstitutionProcessor`, bringing coverage from 60% to 100%.

Tests added:

`test_add_institution` — verifies that the Institution column is correctly populated based on each PI's email domain
`test_add_institution_missing_pi `— verifies that rows where Manager (PI) is missing are skipped without raising an error

Also adds `new_add_institution_processor()`  to `util.py `to follow the existing pattern used by all other processors.


process_report/tests/unit/processors/test_add_institution_processor.py
```diff
@@ -0,0 +1,60 @@
...
+
+        test_data = self._get_test_data(
+            pi_names=["pi1@bu.edu", "pi2@mit.edu"],
+            projects=["ProjectA", "ProjectB"],
+        )
+        processor = test_utils.new_add_institution_processor(data=test_data)
+        processor.process()
+        output = processor.data
+        assert output.loc[0, "Institution"] == "Boston University"
+        assert output.loc[1, "Institution"] == "MIT"
```
QuanMPhm: This feedback applies to the second test case as well. I would instead check the entire dataframe, as we do in most other test cases, like this:


```suggestion
        answer_data = self._get_test_data(
            pi_names=["pi1@bu.edu", "pi2@mit.edu"],
            projects=["ProjectA", "ProjectB"],
            institutions=["Boston University", "MIT"],
        )
        assert output.equals(answer_data)
```

You'll need to edit `_get_test_data` accordingly

QuanMPhm (changes requested):
Just one comment

process_report/tests/unit/processors/test_add_institution_processor.py
```diff
@@ -0,0 +1,64 @@
+from unittest import TestCase, mock
+import pandas
+
+from process_report.tests import util as test_utils
+from process_report.institute_list_models import InstituteList
+
+
+class TestAddInstitutionProcessor(TestCase):
+    def _get_test_data(self, pi_names, projects=None, institutions=None):
```
QuanMPhm: Since you're always passing in `projects`, you might as well remove the default option from this function. Adding the default `None` here suggests the `projects` parameter was optionally in some of your tests, which is not the case

process_report/tests/unit/processors/test_add_institution_processor.py
```diff
@@ -0,0 +1,64 @@
...
+            pi_names=[None, "pi@bu.edu"],
+            projects=["ProjectA", "ProjectB"],
+        )
+
+        processor = test_utils.new_add_institution_processor(data=test_data)
+        processor.process()  # should not raise
+        output = processor.data
+
+        assert pandas.isna(output.loc[0, "Manager (PI)"])
+        assert output.loc[1, "Institution"] == "Boston University"
```
QuanMPhm: Why not do assert equal on the whole dataframe, like you did in the first test?

QuanMPhm (changes requested):
Two comments, and can you squash the commits of the PR afterwards?

process_report/tests/unit/processors/test_add_institution_processor.py
```diff
@@ -0,0 +1,69 @@
...
+        )
+        processor = test_utils.new_add_institution_processor(data=test_data)
+        processor.process()
+        output = processor.data
+        answer_data = self._get_test_data(
+            pi_names=["pi1@bu.edu", "pi2@mit.edu"],
+            projects=["ProjectA", "ProjectB"],
+            institutions=["Boston University", "MIT"],
+        )
+        answer_data = answer_data.astype(output.dtypes)
```
QuanMPhm: Interesting. We usually don't need to cast the type of the dataframe to pass the `equals()` assertion. Especially with just plain `str` dtypes. Did you encountered an error without the casting?

marcoagonzales007: I added it because I wasn't sure if the types would match during the comparison. I'll remove it and verify that the tests still pass without it.

QuanMPhm (approved):
Looks good!

PR #300, opened 2026-06-01
Add unit tests for ColdfrontFetchProcessor

marcoagonzales007:
Part of CCI-MOC/MOC-issues#147 

Adds additional unit tests for ColdfrontFetchProcessor bringing coverage from 76% to 82%.

Tests added:

`test_get_coldfront_api_data_with_filepath` — verifies that when a local filepath is provided data is read from the file instead
`test_get_allocation_data_missing_key` — verifies that allocation entries with missing keys are skipped

### DIFF
--- a/process_report/tests/unit/processors/test_coldfront_fetch_processor.py
+++ b/process_report/tests/unit/processors/test_coldfront_fetch_processor.py
@@ -1,7 +1,7 @@
 from unittest import TestCase, mock
 import pandas
 import pytest
-
+import json
 from process_report.tests import util as test_utils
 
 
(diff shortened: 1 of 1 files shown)


process_report/tests/unit/processors/test_coldfront_fetch_processor.py
```diff
@@ -259,3 +259,31 @@ def test_missing_project_cluster_tuples(self, mock_get_allocation_data):
...
+            proc = test_utils.new_coldfront_fetch_processor(
+                coldfront_data_filepath="fake/path.json"
+            )
+            result = proc._get_coldfront_api_data()
+            assert result == mock_api_data
+
+    @mock.patch(
+        "process_report.processors.coldfront_fetch_processor.ColdfrontFetchProcessor._fetch_coldfront_allocation_api",
+    )
+    def test_get_allocation_data_missing_key(self, mock_get_allocation_data):
```
QuanMPhm: Looking at this test, it does not seem like you need any mocking at all. The test itself can be shortened.

```python
    def test_get_allocation_data_missing_key(self):
        """Malformed allocation entries with missing keys are skipped."""
        test_allocation_data = [
            {
                "resource": {"name": "stack"},
                "project": {"pi": "PI1"},
                "attributes": {},
            },  # missing Allocated Project ID
        ]
        proc = test_utils.new_coldfront_fetch_processor()
        result = proc._get_allocation_data(test_allocation_data)
        assert result == {}
```

QuanMPhm (changes requested):

QuanMPhm (approved):
@knikolla @naved001 Looks good to me

PR #302, opened 2026-06-01
Use coverage to measure code coverage in unit and e2e tests

QuanMPhm:
As part of CCI-MOC/MOC-issues#147. I've added usage of `coverage` with minimum coverage goals. I expect these goals to be increased after we merge in the unit tests that @marcoagonzales007 is writing for the mentioned issue.

### DIFF
--- a/.github/workflows/e2e-tests.yaml
+++ b/.github/workflows/e2e-tests.yaml
@@ -29,6 +29,8 @@ jobs:
 
     - name: Run E2E Tests
       run: |
-        pytest process_report/tests/e2e/test_e2e_pipeline.py -v -s --log-cli-level=INFO
+        coverage run --source="." -m pytest process_report/tests/e2e/test_e2e_pipeline.py -v -s --log-cli-level=INFO
+        coverage combine
+        coverage html --fail-under=85 --omit "process_report/tests/**,process_report/util.py,process_report/institute_list_validate.py"
       env:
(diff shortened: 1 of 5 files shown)


.github/workflows/e2e-tests.yaml
```diff
@@ -29,6 +29,8 @@ jobs:
 
     - name: Run E2E Tests
       run: |
-        pytest process_report/tests/e2e/test_e2e_pipeline.py -v -s --log-cli-level=INFO
+        coverage run --source="." -m pytest process_report/tests/e2e/test_e2e_pipeline.py -v -s --log-cli-level=INFO
+        coverage combine
+        coverage html --fail-under=85 --omit "process_report/tests/**,process_report/util.py,process_report/institute_list_validate.py"
```
larsks, line 34: Rather than running `coverage` explicitly, you can install the [pytest-cov](https://pypi.org/project/pytest-cov/) plugin to have `coverage` run as part of your `pytest` invocation.

larsks (approved):

knikolla (approved):

.github/workflows/e2e-tests.yaml
QuanMPhm: I'm having trouble trying to get `pytest-cov` to properly cover the e2e test. It is not covering the subprocess created during the test. I'll just merge this PR as it is for now. I may look at using `pytest-cov` in the future

PR #306, opened 2026-06-15
Add royalty mechanism

### LINKED ISSUE #136: Implement processor and invoice to show percentage royalties for externally funded projects
## Motivation
We would like to generate a report containing percentage royalties on projects meeting certain criterias:
- Projects whose PIs belongs to "external" institutions (TBA)
- Porjects whose PIs are internal, but have "external funding". Info on whether a project has external funding will be gathered from Coldfront, as through this [planned change](https://github.com/nerc-project/coldfront-plugin-cloud/pull/319)

## Completion Criteria
Processor and invoice implemented, and able to use during monthly invoicing.

## Description
- [ ] Update the institute list models to add a new field for marking external or internal institutions.
- [ ] Create a processor and invoice to process and generate reports for arbitrary percentage royalties

## Completion dates
Desired - 2026-06-20
Required - 2026-07-20

CC @joachimweyl @knikolla @naved001

QuanMPhm:
Closes CCI-MOC/MOC-issues#136. I'll set this as draft, as I have some questions.

This involved several changes:
- New field `moc_member` in institution list
- New pipeline setting `royalty_rate`, default to fetching from nerc_rates
- Coldfront processor will now obtain info on whether a project is externally funded from Coldfront API
- New processor and invoice to calculate and export the royalties

### DIFF
--- a/process_report/invoices/invoice.py
+++ b/process_report/invoices/invoice.py
@@ -76,6 +76,7 @@ CREDIT_FIELD = "Credit"
 CREDIT_CODE_FIELD = "Credit Code"
 SUBSIDY_FIELD = "Subsidy"
 BALANCE_FIELD = "Balance"
+ROYALTY_FIELD = "Royalty"
 ###
 
 ### Internally used field names
@@ -86,6 +87,7 @@ PROJECT_NAME_FIELD = "Project"
 GROUP_MANAGED_FIELD = "MGHPCC Managed"
 CLUSTER_NAME_FIELD = "Cluster Name"
 IS_COURSE_FIELD = "Is Course"
+IS_EXTERNALLY_FUNDED_FIELD = "Is Externally Funded"
 ###
 
 ### Initialized Column objects
@@ -142,6 +144,10 @@ CLUSTER_NAME_COLUMN = InvoiceColumn(name=CLUSTER_NAME_FIELD, dtype=STRING_FIELD_
 IS_COURSE_COLUMN = InvoiceColumn(
     name=IS_COURSE_FIELD, dtype=BOOL_FIELD_TYPE, default_value=False
 )
+IS_EXTERNALLY_FUNDED_COLUMN = InvoiceColumn(
+    name=IS_EXTERNALLY_FUNDED_FIELD, dtype=BOOL_FIELD_TYPE, default_value=False
+)
+ROYALTY_COLUMN = InvoiceColumn(name=ROYALTY_FIELD, dtype=BALANCE_FIELD_TYPE)
 ###
 
 
--- a/process_report/invoices/royalty_invoice.py
+++ b/process_report/invoices/royalty_invoice.py
@@ -1,0 +1,20 @@
+from dataclasses import dataclass
+
+import process_report.invoices.invoice as invoice
+
+
+@dataclass
+class RoyaltyInvoice(invoice.Invoice):
+    name: str = "Royalties"
+    export_columns_list = [
+        invoice.PROJECT_FIELD,
+        invoice.PI_FIELD,
+        invoice.CLUSTER_NAME_FIELD,
+        invoice.INSTITUTION_FIELD,
+        invoice.SU_TYPE_FIELD,
+        invoice.BALANCE_FIELD,
+        invoice.ROYALTY_FIELD,
+    ]
+
+    def _prepare_export(self):
+        self.export_data = self.data[~self.data[invoice.ROYALTY_FIELD].isna()]
--- a/process_report/processors/prepayment_processor.py
+++ b/process_report/processors/prepayment_processor.py
@@ -17,7 +17,7 @@ logging.basicConfig(level=logging.INFO)
 
 @dataclass
 class PrepaymentProcessor(discount_processor.DiscountProcessor):
-    IS_DISCOUNT_BY_NERC = True
+    IS_DISCOUNT_BY_NERC = False
     PREPAY_DEBITS_S3_FILEPATH = "Prepay/prepay_debits.csv"
 
     initializes_columns = (
--- a/process_report/tests/unit/processors/test_prepayment_processor.py
+++ b/process_report/tests/unit/processors/test_prepayment_processor.py
@@ -1,9 +1,15 @@
+import pytest
+
 import pandas
 
 from process_report.tests import util as test_utils
 from process_report.tests.base import BaseTestCaseWithTempDir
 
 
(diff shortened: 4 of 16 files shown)


process_report/invoices/invoice.py
```diff
@@ -142,6 +144,10 @@ class InvoiceColumn:
 IS_COURSE_COLUMN = InvoiceColumn(
     name=IS_COURSE_FIELD, dtype=BOOL_FIELD_TYPE, default_value=False
 )
+IS_EXTERNALLY_FUNDED_COLUMN = InvoiceColumn(
+    name=IS_EXTERNALLY_FUNDED_FIELD, dtype=BOOL_FIELD_TYPE, default_value=False
+)  # TODO: We are fine with this default?
+ROYALTY_COLUMN = InvoiceColumn(name=ROYALTY_FIELD, dtype=BALANCE_FIELD_TYPE)
```
QuanMPhm, line 150: @joachimweyl @naved001 @knikolla Are we fine with `False` as the default for this field. This will mean, if a Coldfront project does not specify the `Is Externally Funded` attribute, invoicing will assume it is internally funded.

process_report/invoices/royalty_invoice.py
```diff
@@ -0,0 +1,21 @@
...
+    export_columns_list = [  # TODO: Confirm list of information we want to include in royalty report
+        invoice.INVOICE_DATE_FIELD,
+        invoice.PROJECT_FIELD,
+        invoice.PI_FIELD,
+        invoice.CLUSTER_NAME_FIELD,
+        invoice.INSTITUTION_FIELD,
+        invoice.SU_TYPE_FIELD,
+        invoice.BALANCE_FIELD,
+        invoice.ROYALTY_FIELD,
+    ]
```
QuanMPhm, line 18: @joachimweyl @knikolla @naved001 Are we fine with these sets of columns for now for the royalty invoice? Are there any other info we want to provide to RH?

process_report/processors/prepayment_processor.py
```diff
@@ -17,7 +17,7 @@
 
 @dataclass
 class PrepaymentProcessor(discount_processor.DiscountProcessor):
-    IS_DISCOUNT_BY_NERC = True
+    IS_DISCOUNT_BY_NERC = False
```
QuanMPhm, line 20: This change was to address the fact that prepayments are still subject to the royalty. With this change, the prepay discount will no longer impact the `Balance` column, which we've used to indicate the money we (the MOC) expect to obtain, which is our revenue, which is the actually value subject to the royalty, as opposed to the money PIs are expected to pay (`PI Balance`), which are not subject to the royalty.

This change will impact how the PI-specific invoices are exported, and as mentioned by @knikolla in [slack](https://massopencloud.slack.com/archives/C07TZC9KKNW/p1781540119467029?thread_ts=1781530441.827499&cid=C07TZC9KKNW), we may want to review how we handle these balance columns and refactor them.

@knikolla @naved001 Given that we have no prepayments yet, are we fine with postponing refactoring the balance fields after this PR is merged?

process_report/invoices/invoice.py
knikolla: I'm fine with `False` as a default. As this field only matters for internal projects, the default of which is internal funding.

process_report/invoices/royalty_invoice.py
knikolla: @joachimweyl

process_report/processors/prepayment_processor.py
knikolla: So this change will break prepayments (which we don't have any yet) temporarily to reuse the Balance column?

process_report/invoices/royalty_invoice.py
joachimweyl: Based on what we have currently, yes.

process_report/processors/prepayment_processor.py
QuanMPhm: @knikolla With this change, even if a prepayment fully covers a PI's costs, the PI-specific invoice will look as if their prepayment hasn't been applied (`Balance` will not be $0). If you believe this breaks prepayment, then yes.

knikolla: Balance is what we ask PIs to pay, so if the balance field doesn't take into account prepayment then I do consider it broken.

QuanMPhm:
@knikolla @naved001 I have responded to all comments and all of my questions have been answered. I believe this is ready for review

process_report/invoices/royalty_invoice.py
knikolla: @QuanMPhm since this is confirmed please remove the comment.

process_report/tests/unit/processors/test_prepayment_processor.py
```diff
@@ -106,7 +106,7 @@ def test_one_group_one_project(self):
         answer_invoice["Prepaid Group Used"] = [1000]
         answer_invoice["Invoice Email"] = ["G1@bu.edu"]
         answer_invoice["PI Balance"] = [0]
-        answer_invoice["Balance"] = [0]
+        answer_invoice["Balance"] = [1000]
```
knikolla: You shouldn't change the balance output. Since the feature will be broken after this, you should set the test as expected to fail for the tests that require it. This way when the feature is fixed, we can remove the expectation of failure and restore the correct test behavior.

knikolla (changes requested):
I'm okay with temporarily breaking Prepayments. See one comment, otherwise looks good. Confirm with Kim that breaking prepayments temporarily is okay.

QuanMPhm:
@knikolla I have confirmed with @joachimweyl that we're fine breaking prepayments for now.

process_report/tests/unit/processors/test_prepayment_processor.py
```diff
@@ -1,9 +1,12 @@
+import pytest
+
 import pandas
 
 from process_report.tests import util as test_utils
 from process_report.tests.base import BaseTestCaseWithTempDir
 
 
+@pytest.mark.xfail(strict=True, reason="Test")
```
knikolla: Here you want to add the actual reason as a string, something along the lines of prepayments being temporarily broken to implement the royalty mechanism.

QuanMPhm: Sorry. Forgot to replace the placeholder string during testing

knikolla (approved):

PR #308, opened 2026-06-29
Increase e2e coverage for NewPICredit and Prepayment processors

marcoagonzales007:
Part of CCI-MOC/MOC-issues#147 

Updates the e2e test data so the full pipeline covers previously-uncovered branches in the two processors

`NewPICreditProcessor` (80% -> 90%)
- Changed pi1@bu.edu to a "first invoice month" of 2025-06 (age 0) and pi2@harvard.edu to 2025-05 (age 1) in `test_PI.csv`
- This makes both PIs credit eligible within the age window, so the credit-application logic now is able to run

`PrepaymentProcessor` (82% -> 87%)
- Added a dedicated PI (pi7@mit.edu) and project (P7) whose balance survives the subsidy processors, so the prepayment step applies the discount
- Pointed TestGroup1's prepay project at P7 and extended its end date to 2025-12 so it's active for the invoice month
- This covers the active prepay project logic and the debit-recording path that were never reached before

### DIFF
--- a/process_report/tests/e2e/test_data/test_PI.csv
+++ b/process_report/tests/e2e/test_data/test_PI.csv
@@ -2,5 +2,6 @@ PI,First Invoice Month,Initial Credits,1st Month Used,2nd Month Used
 pi5@harvard.edu,2025-06,0.00,0.00,0.00
 pi6@mit.edu,2025-06,0.00,0.00,0.00
 pi4@example.edu,2025-06,0.00,0.00,0.00
-pi1@bu.edu,2025-04,1000.00,300.00,0.00
-pi2@harvard.edu,2025-04,1000.00,300.00,0.00
+pi1@bu.edu,2025-06,1000.00,320.00,0.00
+pi2@harvard.edu,2025-05,1000.00,300.00,35.00
+pi7@mit.edu,2025-04,1000.00,1000.00,0.00
--- a/process_report/tests/e2e/test_e2e_pipeline.py
+++ b/process_report/tests/e2e/test_e2e_pipeline.py
@@ -117,12 +117,13 @@ def _prepare_pipeline_execution(
     env["FETCH_FROM_S3"] = "false"
     env["UPLOAD_TO_S3"] = "false"
     env["invoice_path_template"] = str(test_files["test_invoice_dir"])
-
-    pi_file_copy = workspace / "test_PI.csv"
+    input_dir = workspace / "input_data"
+    input_dir.mkdir(exist_ok=True)
+    pi_file_copy = input_dir / "test_PI.csv"
     shutil.copy(test_files["test_PI.csv"], pi_file_copy)
     env["PI_REMOTE_FILEPATH"] = str(pi_file_copy)
     env["ALIAS_REMOTE_FILEPATH"] = str(test_files["test_alias.csv"])
-    prepay_debits_copy = workspace / "test_prepay_debits.csv"
+    prepay_debits_copy = input_dir / "test_prepay_debits.csv"
     shutil.copy(test_files["test_prepay_debits.csv"], prepay_debits_copy)
     env["PREPAY_DEBITS_REMOTE_FILEPATH"] = str(prepay_debits_copy)
     env["PREPAY_CREDITS_FILEPATH"] = str(test_files["test_prepay_credits.csv"])
--- a/process_report/tests/e2e/test_e2e_pipeline.py
+++ b/process_report/tests/e2e/test_e2e_pipeline.py
@@ -1,4 +1,5 @@
 import os
+import shutil
 from pathlib import Path
 import pandas as pd
 import pytest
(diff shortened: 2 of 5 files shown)


process_report/tests/e2e/test_data/test_PI.csv
```diff
@@ -2,5 +2,6 @@ PI,First Invoice Month,Initial Credits,1st Month Used,2nd Month Used
 pi5@harvard.edu,2025-06,0.00,0.00,0.00
 pi6@mit.edu,2025-06,0.00,0.00,0.00
 pi4@example.edu,2025-06,0.00,0.00,0.00
-pi1@bu.edu,2025-04,1000.00,300.00,0.00
-pi2@harvard.edu,2025-04,1000.00,300.00,0.00
+pi1@bu.edu,2025-06,1000.00,320.00,0.00
```
QuanMPhm, line 5: Based on your PR description and the changes you made, it seems that since the e2e test will now actually apply the New-PI credit and prepayments to some invoices, mutable test files (`test_PI.csv` and `test_prepay_debits.py`) will now be changed after each run of the e2e test. Could you edit the code in `test_e2e_pipeline.py` so that the aforementioned files are not changed after each run? My suggestion comes from the idea that what we commit into the repo should only be the test cases, not their results or side effects. Investigate the ways you can accomplish this, and let me know what solution you'll like to implement.

Also, I'm curious why I don't see changes to `test_prepay_debits.py` in this PR diff. Was it not changed when you ran the e2e test on your machine?

QuanMPhm (changes requested):
Just one comment

process_report/tests/e2e/test_data/test_PI.csv
marcoagonzales007: Thank you for your feedback, and I think It was changed by the pipeline itself during the run. I reverted it thinking it was a stray change, but I now realize that's the side effect of the problem you have just described. I'll work on `test_e2e_pipeline.py` so that the files are not changed after each run. my plan for this is to copy the mutable files into the temp workspace and point the pipeline at these copies, so the test data stays untouched. Let me know if that sounds good. Thank you.

QuanMPhm: > I'll work on test_e2e_pipeline.py so that the files are not changed after each run. my plan for this is to copy the mutable files into the temp workspace and point the pipeline at these copies,

Nice! That was the solution I had in mind too. Let me know when you've amended this PR.

process_report/tests/e2e/test_e2e_pipeline.py
```diff
@@ -116,11 +117,15 @@ def _prepare_pipeline_execution(
     env["FETCH_FROM_S3"] = "false"
     env["UPLOAD_TO_S3"] = "false"
     env["invoice_path_template"] = str(test_files["test_invoice_dir"])
-
-    env["PI_REMOTE_FILEPATH"] = str(test_files["test_PI.csv"])
+    input_dir = workspace / "input_data"
```
QuanMPhm, line 120: Is there a reason we want to create an additional folder here just for the mutable data files? What would be the benefit compared to just copying them into `workspace`?

QuanMPhm (commented):
One comment

process_report/tests/e2e/test_e2e_pipeline.py
marcoagonzales007: Good question. The subfolder is there because of `_validate_outputs()`, it runs workspace.glob("*.csv") and fails the test on any unexpected CSV in the workspace root. If I copy test_PI.csv and test_prepay_debits.csv into the root directly, then they would get flagged as "unexpected CSV files", that was the CI failure I hit before adding the subfolder. Putting the copies in input_data/ keeps them out of that top-level scan while still letting the pipeline read/write them.

I'll make sure to add a comment explaining this so it's clear in the code, or If you'd rather keep everything in the workspace root, I could exclude those two filenames from the validation check instead. Happy to go whichever route.

QuanMPhm: Make sense. Now that I see this, I would have preferred if you just made another tempdir, specifically for storing mutable input files, instead of shoving everything into the same folder. This is fine for now. If the use of the `workspace` becomes more conflated in the future, I will ask that you divide the folder up

process_report/tests/e2e/test_data/test_PI.csv
```diff
@@ -2,5 +2,6 @@ PI,First Invoice Month,Initial Credits,1st Month Used,2nd Month Used
...
+pi2@harvard.edu,2025-05,1000.00,300.00,35.00
```
QuanMPhm: nitpick: Your PR increases coverage by changing the age of the PIs. But here, you are also changing the credits they've used. These seem to be changes unrelated to the PR

QuanMPhm (approved):
Looks good. Just some minor feedback you don't have to implement here. @knikolla @naved001 Off to you guys

PR #309, opened 2026-07-06
Add standard invoice CSV reader/writer

marcoagonzales007:
Part of CCI-MOC/MOC-issues#139 

Adds a standard reader/writer (`invoice_csv.py`) so all invoice csv's are read and written with one format:

quote character: |
quoting: only fields containing special characters get quoted
read engine: pyarrow

**Changes**: 

1. New `process_report/invoice_csv.py` with `write_invoice_csv()` and `read_invoice_csv()`
2. `Invoice.export()` now writes through the standard writer
3. `merge_csv()` now reads through the standard reader 
4. Unit tests covering round-trip, fields containing commas, and that the standard quote char is actually used

### DIFF
--- a/process_report/process_report.py
+++ b/process_report/process_report.py
@@ -3,7 +3,7 @@ import logging
 import os
 
 import pandas
-
+from process_report import invoice_csv
 from process_report.settings import invoice_settings
 from process_report.loader import loader
 from process_report import util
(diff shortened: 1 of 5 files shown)


process_report/process_report.py
```diff
@@ -106,9 +106,8 @@ def merge_csv(files):
     """Merge multiple CSV files and return a single pandas dataframe"""
     dataframes = []
     for file in files:
-        dataframe = pandas.read_csv(
+        dataframe = invoice_csv.read_invoice_csv(
             file,
-            engine="pyarrow",
             dtype={
```
QuanMPhm, line 111: AFAIK, there's no reason why the `dtype` argument here would ever change. Please move this dtype dict into `invoice_csv.py`, and remove `dtype` as an argument from `read_invoice_csv`.

This will have implications for your test cases, since the reader will automatically cast the dtype of certain columns.

QuanMPhm (changes requested):

PR #310, opened 2026-07-09
Enable exporting monthly invoice to Iceberg

### LINKED ISSUE #144: Feature: Append dataframe to Iceberg during invoicing pipeline run

jimmysway:
- Takes over [#292](https://github.com/CCI-MOC/invoicing/pull/292) from @QuanMPhm so we can continue development from this fork.
- Rebased Quan's Iceberg export commit onto current `main` (column-tracking from CCI-MOC/invoicing#285 is already merged).
- Next step: rewrite the export path to use PySpark instead of pyiceberg, because pyiceberg does not support Hadoop catalogs ([apache/iceberg-python#17](https://github.com/apache/iceberg-python/issues/17)), which is what our S3 Iceberg table uses.

Closes CCI-MOC/MOC-issues#144.

### DIFF
--- a/process_report/invoices/iceberg_invoice.py
+++ b/process_report/invoices/iceberg_invoice.py
@@ -1,0 +1,82 @@
+import logging
+from dataclasses import dataclass, field
+
+from pyiceberg.table import Table
+from pyiceberg.catalog import Catalog, load_catalog
+import pyarrow
+
+import process_report.invoices.invoice as invoice
+from process_report.loader import loader
+from process_report.settings import invoice_settings
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+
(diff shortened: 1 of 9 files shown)


jimmysway:
Will push directly onto Quan's fork

**Edit**: Didn't have permission so I will start the new pyspark writing implementation here

